PR #246, opened 2016-10-04
Build with AppVeyor CI

jhasse:
This uses [MSYS2](https://msys2.github.io/) to compile, which includes wxWidgets. It even has premake, but for now it's still use the .bat file and therefore the premake4.exe included in git.

The wx-config binary included in wxWidgets is using --rescomp instead of --rcflags. `wx-config --rescomp` returns the full compiler command, e. g.

```sh
$ wx-config --rescomp
windres --include-dir C:/msys64/mingw32/include/wx-3.0 --define __WIN32__ --define __GNUWIN32__
```

That's why I added the `| cut -d' ' -f2-`, to remove the "windres".

These lines

```sh
ln -s /mingw32/include/binutils/bfd.h /mingw32/include/bfd.h
ln -s /mingw32/include/binutils/symcat.h /mingw32/include/symcat.h
ln -s /mingw32/lib/binutils/libbfd.a /mingw32/lib/libbfd.a
ln -s /mingw32/lib/binutils/libiberty.a /mingw32/lib/libiberty.a
```

are needed because MSYS2 adds those headers and libs in a subfolder, I don't know why (I've asked on the msys2 mailing list [here](https://sourceforge.net/p/msys2/mailman/message/35406062/)).

```
sed 's!\$(LDFLAGS) \$(RESOURCES) \$(ARCH) \$(LIBS)!\$(LIBS) \$(LDFLAGS) \$(RESOURCES) \$(ARCH)!g' *.make -i"
```
Somehow MinGW doesn't like that wxWidgets' libs come before static libs. That's why I swap LDFLAGS and LIBS here.

```
sed 's!-lbfd!-lbfd -lz!g' *.make -i"
```
bfd.a needs zlib. Maybe it can be added in the lua scripts in the future, but I didn't want to mess with them for now.

Besides these few hacks it works fine, you can see the result here: https://ci.appveyor.com/project/jhasse/wxformbuilder

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,31 @@
+version: 1.0.{build}
+environment:
+  MSYSTEM: MINGW32
+install:
+- cmd: >-
+    cinst -y InnoSetup
+
+    C:\msys64\usr\bin\bash -lc "pacman -S --needed --noconfirm mingw-w64-i686-wxWidgets"
+build_script:
+- cmd: >-
+    git submodule update --init
+
+
+    create_build_files4.bat --wx-root=/mingw32/bin --force-wx-config --disable-mediactrl
+
+
+    C:\msys64\usr\bin\bash -lc "ln -s /mingw32/include/binutils/bfd.h /mingw32/include/bfd.h"
+
+    C:\msys64\usr\bin\bash -lc "ln -s /mingw32/include/binutils/symcat.h /mingw32/include/symcat.h"
+
+    C:\msys64\usr\bin\bash -lc "ln -s /mingw32/lib/binutils/libbfd.a /mingw32/lib/libbfd.a"
+
+    C:\msys64\usr\bin\bash -lc "ln -s /mingw32/lib/binutils/libiberty.a /mingw32/lib/libiberty.a"
+
+
+    C:\msys64\usr\bin\bash -lc "cd '%cd%'/build/3.0/gmake && sed 's!\$(LDFLAGS) \$(RESOURCES) \$(ARCH) \$(LIBS)!\$(LIBS) \$(LDFLAGS) \$(RESOURCES) \$(ARCH)!g' *.make -i"
+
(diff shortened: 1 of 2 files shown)


rjpcomputing:
That is great work!

I know that you are putting in a bunch of effort and time. I think you should just make these changes and not wait for me. I am more of a user these days than a developer. What do you think?

Also I have wxFB building on an internal CI and I am afraid that the change to wxwidgets.lua will break that. Is there any reason that --rcflags doesn't work. I think I just don't understand AppVeyor and it's build environment enough. Is there an alternative when building on AppVeyor? Is using MSYS2 the way we want to have all Windows developers build wxFB? Thoughts?

jhasse:
> That is great work!

Thanks :)

> I know that you are putting in a bunch of effort and time. I think you should just make these changes and not wait for me. I am more of a user these days than a developer. What do you think?

Sounds good, I will  merge this PR then and we can use it for further discussions about AppVeyor / MSYS2.

> Also I have wxFB building on an internal CI and I am afraid that the change to wxwidgets.lua will break that. Is there any reason that --rcflags doesn't work.

That's why I did it like this:

```
"`" .. configCmd .." --rcflags || " .. configCmd .." --rescomp | cut -d' ' -f2-`"
```

This way it will still try `--rcflags` first and only if it failed try again with `--rescomp`. This should still work with your internal CI, can you try?

> Is there an alternative when building on AppVeyor?

Yeah, it's possible to create a local `wx-config` script which changes the flags on the fly.

> Is using MSYS2 the way we want to have all Windows developers build wxFB? Thoughts?

At least for MinGW builds it should be, as it makes many things so much easier (like installing dependencies). It also comes with the latest MinGW-w64 and it's pre-installed on AppVeyor.

Visual C++ is a different story though ;)

rjpcomputing:
@jhasse 

The builds worked great. Great work.

Visual C++ is not currently supported and has not ever been.

patlkli:
When the work on the CMake build system is finished, Visual C++ will also be supported.

Issue #262, opened 2016-12-14
Create Windows installer using AppVeyor


jhasse:
* [x] Run InnoSetup to create the installer
* [ ] Automatically include all the needed DLL files
* [ ] Upload the installer automatically to a GitHub release

rugruth:
Simply Awesome. 

Keep up the good work :)

R.

jhasse:
Okay AppVeyor now builds the installer. If someone wants to try it out: https://ci.appveyor.com/api/buildjobs/h066tldremgkpfed/artifacts/install%2Fwindows%2FwxFormBuilder_v3.6.0.exe

If just included all the DLLs like it was done in #263. The installer is only 11 MB, I think that's okay.

We'll have to upload it to a GitHub release manually for now, I couldn't find a way to do this without also creating the release via AppVeyor. Ideally I would want AppVeyor only to build the installer when I push a tag.

TcT2k:
From what I understood from the AppVeyor documentation on [github releases](https://www.appveyor.com/docs/deployment/github/), you can limit the deployment to tag pushes only (see last line in the sample config).

PR #263, opened 2016-12-14
Upload Win32 build as appveyor artifact to allow downloading binaries

TcT2k:
Package a successful appveyor build as an artifact and add a download link to the README file to allow easy access to current Windows binaries.

This could be expanded to include setups, if they would be built in the appveyor script.


jhasse:
Thank you! I've just create #262 btw ;)

PR #265, opened 2016-12-16
CMake build fixes for windows

TcT2k:
* Fixes various small issues to enable building with visual studio
* Removed dependency on boost (from @beached's fork)
* Fix to build with 3.1.1
* Modified appveyor
  * build with VS2013 using wxWidgets 3.0.2
  * build with VS2015 usign wxWidgets 3.1.0
  * build using mingw

The mingw build currently fails when building stack.cpp and I could use some help what is missing there. All packages should be installed as the premake build worked.


patlkli:
Thank you!

I think we have to add the binutils subdirectory to the include path for the MinGW build.
Can you do that and check if it builds for you then?

TcT2k:
The binutils include path and linking the correct libraries seems to do the trick.

PR #268, opened 2017-01-09
Add sizer flag wxRESERVE_SPACE_EVEN_IF_HIDDEN.

### LINKED ISSUE #93: Doesn't know of wxRESERVE_SPACE_EVEN_IF_HIDDEN flag
Tried to add wxRESERVE_SPACE_EVEN_IF_HIDDEN sizeritembase flag property

A message ‘! flag: Unknown flag identifier “wxRESERVE_SPACE_EVEN_IF_HIDDEN”’ is shown in the status bar

This flag is relatively new (Available from wx version 2.8.8)

Reported by: jfouche

Poddster:
No idea if this is the correct way to add a flag (I've read neither the documentation nor the codebase), but it seems to work on a locally built wxFormBuilder.

Will fix #93.

http://docs.wxwidgets.org/trunk/classwx_sizer.html

>wxRESERVE_SPACE_EVEN_IF_HIDDEN 	
>Normally wxSizers don't allocate space for hidden windows or other items. This flag overrides this behaviour so that sufficient space is allocated for the window even if it isn't visible. This makes it possible to dynamically show and hide controls without resizing parent dialog, for example. (Available since 2.8.8.)


patlkli:
That's exactly the way to do it.
Thank you!

PR #276, opened 2017-02-14
Automatic creation of CustomControl declaration and construction

jgonzalezdr:
For easier custom control usage and management from wxFormBuilder, the declaration and construction of custom controls can be created with default values, which will probably be just fine in most cases.


jhasse:
Thanks :)

PR #281, opened 2017-04-24
Fix for 2 wxPyDeprecationWarning with the latest wxPython==4.0.0a1

a-atalla:
A fix for the following deprecation warning  in the generated python code, may be related to 
this [issue](https://github.com/wxFormBuilder/wxFormBuilder/issues/275)

``` 
wxPyDeprecationWarning: Call to deprecated item. Use Append instead.
  self.m_menu1.AppendItem( self.m_menuItem1 )
```

And 

```
wxPyDeprecationWarning: Call to deprecated item. Use SetSizeHints instead.
  self.SetSizeHintsSz( wx.DefaultSize, wx.DefaultSize )
```


jhasse:
Thanks :)

PR #304, opened 2017-08-30
Updated 'AddSpacer' to new syntax in Phoenix

### LINKED ISSUE #303: AddSpacer generated code uses incorrect parameters
According to the [wx documentation](https://wxpython.org/Phoenix/docs/html/wx.BoxSizer.html#wx.BoxSizer.AddSpacer), a "BoxSizer"'s 'AddSpacer' function has only one parameter, which is the size of the stretched space. 

In wxFormBuilder, four parameters are inputted; a tuplet set of x and y values, the proportion, whether to expand, and the border size. 

This makes the program refuse to run due to TypeError, at least with Python.

Are there any workarounds (like using another object instead of a spacer)?

EDIT: I can see one (easy) solution - change the 'AddSpacer' function call to instead call 'Add'. Unfortunately, I cannot seem to access 'Add' from within wxFormBuilder.

EDIT 2: My pull request should fix this.

Thanks.

TheDogeOfTheInternet:
Changed 'AddSpacer' to 'Add' as now 'AddSpacer' has only one argument. If it does what I intend for it to do, it should fix #303 and #275.

### DIFF
--- a/output/plugins/layout/xml/layout.pythoncode
+++ b/output/plugins/layout/xml/layout.pythoncode
@@ -38,13 +38,13 @@ Python code generation writen by
   <templates class="sizeritem">
     <template name="window_add">#parent $name.Add( self.#child $name, $proportion, $flag, $border ) #nl</template>
     <template name="sizer_add">#parent $name.Add( #child $name, $proportion, $flag, $border ) #nl</template>
-    <template name="spacer_add">#parent $name.AddSpacer( ( #child $width, #child $height), $proportion, $flag, $border ) #nl</template>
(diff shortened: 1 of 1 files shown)


jhasse:
> If it does what I intend for it to do

Have you tested it?

TheDogeOfTheInternet:
I've tested what I set the output to in Python, and it runs without error.

However, I have not tested the compiled program. I'll test now.

TheDogeOfTheInternet:
Yeah, works perfectly.

jhasse:
Thank you :)

btw: For Git commit messages try to use the imperative mood in the subject line (Update instead of Updated).

PR #314, opened 2017-10-05
Syntax fixes for SetToolTipString and wxST_SIZEGRIP

jmagic:
I've made a couple changes to correct errors I was getting when using the generated python code. The python code works correctly now, but I'm not sure how to run the built in tests, or if this will effect other code generation. 

Warnings received:
wxPyDeprecationWarning: Call to deprecated item. Use SetToolTip instead.

AttributeError: module 'wx' has no attribute 'ST_SIZEGRIP'

Thanks,
Jim

### DIFF
--- a/output/plugins/common/xml/menutoolbar.xml
+++ b/output/plugins/common/xml/menutoolbar.xml
@@ -27,7 +27,7 @@ Written by
     <property name="name" type="text">m_statusBar</property>
     <property name="fields" type="uint"  help="Number of fields in the statusbar.">1</property>
     <property name="style" type="bitlist">
-      <option name="wxST_SIZEGRIP"       help="On Windows, displays a gripper at right-hand side of the status bar." />wxST_SIZEGRIP</property>
+      <option name="wxSTB_SIZEGRIP"       help="On Windows, displays a gripper at right-hand side of the status bar." />wxSTB_SIZEGRIP</property>
   </objectinfo>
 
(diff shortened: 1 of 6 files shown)


jhasse:
Thanks :)

Alchemist45:
This is apparently not in the wxFormBuilder 3.6.2 version I downloaded.  Is there a newer one and where can I find it?

jhasse:
It should be in 3.6.1 and therefore 3.6.2.

Alchemist45:
Sorry to say it is not.  The references to STB_ are to the old ST_ for status bars.  How can I fix this?

Scott

Here is what I got for a form.

self.m_statusBar1 = self.CreateStatusBar( 1, wx.ST_SIZEGRIP|wx.SIMPLE_BORDER, wx.ID_ANY )

When I changed to wx.STB_SIZEGRIP it worked.


-----Original Message-----

jmagic:
Hi Alchemist45,

Could you send a screenshot showing the issue like this:

image


This way I'll see if I can recreate the issue.

Thanks,
Jim M

Alchemist45:
I think I found the problem.  You have to find the menubar style and select wxSTB_SIZEGRIP or the program selects the old wxST_SIZEGRIP which is no longer permitted.
Once I selected that option it worked.   Don't know why the old style is the default.

Scott

-----Original Message-----

PR #319, opened 2017-10-16
Update for moved modules in phoenix 

jmagic:
Hi,

I've updated the moved modules in phoenix for python code generation, as per ase1590 #275 request.  I went through all the classes listed here: https://wxpython.org/Phoenix/docs/html/classic_vs_phoenix.html#modules-which-have-moved

Thanks,
Jim M


jhasse:
Thanks :)

PR #321, opened 2017-10-19
Added the ability to change if a subclass is forward declared or the header is included

csteenwyk:
Added the ability to change if a subclass is forward declared or the header is included in the generated file. The default is to forward declare it (previous behavior). This included allowing a child property to have a type. Currently only PT_BOOL and PT_WXSTRING are supported

### DIFF
--- a/output/xml/default.xml
+++ b/output/xml/default.xml
@@ -130,6 +130,7 @@
     <property name="subclass" type="parent" help="For easy use of custom widgets which are simple variations from standard ones, without requiring a new plugin for wxFB or a full xrc handler for wxWidgets. For C++, this replaces the name of the class. For XRC, this sets the subclass value on the object tag.">
 		<child name="name" help="The name of the subclass."/>
 		<child name="header" help="For C++ Only. The header to be included for the subclass."/>
+		<child name="forward_declare" type="bool" help="For C++ Only. Forward declare the subclass, otherwise include the header.">forward_declare</child>
 	</property>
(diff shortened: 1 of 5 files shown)


jhasse:
I think currently wxFormBuilder forward declares *and* includes the header. With your PR it only does one of these things. Am I missing something?

jhasse:
Ah sorry, I've missed https://github.com/wxFormBuilder/wxFormBuilder/commit/81a138b1dead3a3aa054d2753804591c62544350

With that commit you can disable the forward declaration by specifying a header. What's the use-case for a dedicated setting?

ryanjmulder:
wxFormBuilder prefers to forward declare the subclass in the generated .h file, and include the subclass's header in the generated .cpp file. 

Sometimes a forward declaration is not enough, because C++ requires the full definition of the subclass. For example, this happens when subclassing a Frame component, because the generated class for the Frame inherits from the subclass, and the full subclass definition is required for inheritance. 

In those cases, wxFormBuilder rightly decides to include the subclass's header in the generated .h file (instead of the generated .cpp file) which makes the forward declaration redundant. I noticed that and removed the redundant forward declaration in 81a138b.

wxFormBuilder has always had conflict about whether to forward declare the subclass or not. I introduced the current logic in dbbd4d9, but even after that, there have been users that preferred to include the subclass's header in the generated .h file. 

I think allowing the user to decide according to their code's needs or their preferences ultimately makes the most sense, so I like this change a lot.

PR #323, opened 2017-10-27
Fixed typo in "EVT_GRID_CELL_CHANGED" in Python and Cpp code plugins

solidxsnake:


jhasse:
Actually not a typo, but the constant name was changed in wxWidgets 2.8 or 3.0 I think. Thanks for the PR :)

PR #325, opened 2017-11-03
Add icon for wxSpinCtrlDouble

NewPagodi:
When wxSpinCtrlDouble was added, the icon seems to have been missed. This just adds a simple icon that's the same as the spin control icon except it says "1.1" instead of "10". Here's a what it looks like in png form:
spin_ctrl_double


jhasse:
Shouldn't the icon also be used somewhere?

NewPagodi:
The xpm file was already referenced in output\plugins\additional\xml\additional.xml at line 277:
`<objectinfo class="wxSpinCtrlDouble" icon="spin_ctrl_double.xpm" type="widget">`
but there is no such file in the currently in the wxFormBuilder repo.  Consequently, the tool in the toolbar shows up as with a question mark icon.

Because of the line in the xml file mentioned above, when the spin_ctrl_double.xpm file from this pull request is added to the repo, it will be used correctly by wxFormBuilder.

jhasse:
I see! Thank you very much :)

PR #335, opened 2017-12-06
Python CG: Add missing self before instance name of TreeListCtrl when appending

UsYer:


jhasse:
Hm ... any idea why travis fails on macOS?

UsYer:
I think there is a problem in the mac build script: 
> 3.33s$ brew upgrade boost
> Error: boost 1.65.1 already installed
> The command "brew upgrade boost" exited with 1.

This is what sets the overall build status to 1. All other steps seem to work fine. I never worked with travis before. Would you be able to adjust the script so that an already installed dependency does not lead to an error?

UsYer:
Okay, I found a solution for the mac build problem. Please see PR https://github.com/wxFormBuilder/wxFormBuilder/pull/337

UsYer:
@jhasse can you merge these PRs? All jobs are passing.

PR #337, opened 2017-12-06
Don't fail if boost dependency is already up to date

UsYer:
Adapted from: https://docs.travis-ci.com/user/reference/osx/#A-note-on-upgrading-packages


jhasse:
Thanks :)

PR #344, opened 2018-01-10
Fix application bundle creation on macOS to produce usable application

### LINKED ISSUE #340: App completely crashes on macOS
# How to reproduce the crash
- I have tried using the precompiled .zip from release v3.6.1 - app opens and immediately shuts down.
- I have tried to build wxFormBuilder myself following instructions in README.md - app crashes with following crash log: https://gist.github.com/taniotanio7/d6b611619a2478f5cf619a77d8b3291f
- Finally... I have tried to use solution mentioned in #313 (building with debug enabled) - same results as if using precompiled version

# The workaround
Currently, the only solution to make the app work was to use the old version 3.5.01 hosted on [SourceForge](https://sourceforge.net/projects/wxformbuilder/)

nohal:
Fixes #340 
With these changes, the app bundle produced from travis builds should be actually usable without any manual modifications.
Might also be a fix for #313, although there is no useful info there.


jhasse:
Thanks! The Linux failure on Travis is unrelated.

Issue #27, opened 2011-02-11
Python Options (tabs-spaces)


wxfbint:
It would be nice to have an option in the Python project options to select whether one prefers tabs or 4 spaces for the source output.  Keep up the great work!

Reported by: the_bob

Tirpitz93:
bump

awinner1:
Can I bump this, would like to have the option honored for the inherited class(s) as well.....

PR #350, opened 2018-02-03
Python Options (tabs-spaces) #27

ANoDE85:
This commit is to address issue #27 

It's a rather simple fix, but I'm not sure if I thought of everything.


jhasse:
Thanks :)

Issue #62, opened 2015-05-14
Add support for the wxTimePickerCtrl to wxFormBuilder


wxfbint:
Apologies if this is the wrong way to do this.
I am using 3.5 and couldn't find the wxTimePickerCtrl ( I hope I am not just being blind )
So I decided to add it. Here is the patch I did. I haven't included the icon for tool bar.
Great Tool by the way. Thanks.

Reported by: unholytony

PR #351, opened 2018-02-05
Implement wxTimePickerCtrl support

nohal:
This is an old patch picked up from [SF](https://sourceforge.net/p/wxformbuilder/feature-requests/62/) adding support for the [wxTimePickerCtrl](http://docs.wxwidgets.org/trunk/classwx_date_picker_ctrl.html) control.
The only change so far is adding the icon missing in the original patch.

Implements #62


jhasse:
Thanks :)

PR #352, opened 2018-02-05
Unhardcode icon file type for GNU/Linux launcher

palob:
Since you install the icon to a location compliant with [freedesktop.org standards](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) (``usr/share/pixmaps``) you don't need to specify the icon file extension in the `.desktop` launcher. It will be found anyway.

This facilitates the use of non-PNG (e.g. SVG) app icon themes.
Cf. this [example `.desktop` launcher](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#example).


jhasse:
flatpak-builder doesn't seem to like it: 

```
Error loading desktop file: Failed to find icon: Failed to find icon wxformbuilder
```
https://travis-ci.org/wxFormBuilder/wxFormBuilder/jobs/337753457

palob:
Maybe the icon should rather be installed to ``usr/share/icons/hicolor/<size>/apps`` which is the no.1 standard location. I haven't delved into Flatpak yet myself, @bil-elmoussaoui any idea what's the issue here?

bil-elmoussaoui:
@palob  I will take a look soon!

bil-elmoussaoui:
The icon name used on the desktop file doesn't follow the RDNN conventions, so you should use rename-destkop, rename-icon and rename-appdata on the JSON file

Issue #348, opened 2018-01-22
wxFormBuilder stop rendering after changing something in Properties Window on OS X


gamelaster:
I wanted to click EXPAND on element, but after that, the whole application absolutely stop rendering, whatever I click, the application doesn't change.

![GIF](http://g.recordit.co/klbyUQZNVY.gif)

macOS Serria 10.12, wxFormBuilder 3.6.1 (with fixes to make it running)

jhasse:
> with fixes to make it running

Which fixes?

gamelaster:
https://github.com/wxFormBuilder/wxFormBuilder/issues/340#issuecomment-354300996
and
https://github.com/wxFormBuilder/wxFormBuilder/issues/340#issuecomment-354303253

nohal:
This odd behaviour is not new, present even in 3.5 RC1 I still have here - Usually after changing a textual property of a "any" widget in a project that is not totally trivial, the GUI gets mostly unresponsive. I always get to it, but usually not as fast as @gamelaster (What he shows is 100% reproducible here).

jhasse:
Thanks to @nohal this should now be fixed when installing wxWidgets via `brew install wxmac --HEAD`.

Should we update Travis CI or wait for the release of 3.1.1 or even 3.2?

nohal:
I would probably wait for wx 3.1.1 release and then switched travis to use `brew install wxmac --devel` as soon as it means 3.1.1 - should be really soon now.
WX 3.2 can be anything from months to years away and I wouldn't wait for it, at least on macOS where the current wx code is really a big step forward from 3.0.

Also note that all my testing of wxFB built against wx3.1.1-rc was limited to 2 minutes after the initial "Wow, it works..."

jhasse:
Okay :) Btw: we're already using wxWidgets 3.1 for the Linux build. And due to #295 wxWidgets 3.2 can't come fast enough for Windows, too. It's a real pity this is taking so long.

I'll close this issue for now.

PR #353, opened 2018-02-06
Trivial changes to make build against wx3.1.1-rc pass

nohal:
At first glance the result seems pretty usable (And for example #348 seems not to be there anymore...)


nohal:
Have to have a look - I am normally not using Homebrew wxmac at all as I need backward compatibility all the way to OS X 10.7 etc.

nohal:
- `brew install wxmac --devel` - currently at 3.1.0 - Unusable as the property grid is broken. (This will likely get fixed as soon as 3.1.1 replaces it in a few days)
- `brew install wxmac --HEAD` - Same positive result as my local build, #348 seems gone.

FWIW, there is a behavioral change in the property grid - with older versions, hitting Enter in a property grid cell persisted the value and left the cell (And that for me was the usual cause of #348, although not the only one as the video there shows), this is not the case with wx 3.1.1 - hitting Enter in the property grid cell does nothing.

jhasse:
And just `brew install wxmac`?

nohal:
There #348 is present - this is how wxFB 3.6.2 was built (And I had the same result locally at the moment). This PR changes nothing on that (Although it does not break the build against wx 3.0 as we can see on Travis).

jhasse:
Okay. I don't have a Mac so I can't test any of it ;)

But it sounds reasonable, thanks for the fix :)

PR #366, opened 2018-03-07
update version in linux install instructions

doleron:


jhasse:
Not all Linux distributions come with wxWidgets 3.1 yet, so this won't work for them.

doleron:
That's ok.

PR #376, opened 2018-03-29
Fix the wxFBBitmapProperty 'ico_size' values and wxFormBuilder crash when the 'Load From Icon Resource' 'resource_name' property is changed

tomay3000:
Use **Squash and merge** option to exclude the commits history.
[https://help.github.com/articles/about-pull-request-merges/](https://help.github.com/articles/about-pull-request-merges/)


jhasse:
I have no idea what this is about, sorry. Please describe your change a little bit more.

tomay3000:
@jhasse, Take a look at the files changed and ignore the 12 commits.
This is about the main wxPropertyGrid when using ribbon bars under windows, you have the option to load icons from resources.
So when you select '**Load From Icon Resource**' from the dropdown list of the **wxFBBitmapProperty** property category, then some weird values for '**ico_size**' property will be instead of the deafault ones: **(-1; -1)**. and if you try to change these values then wxFormBuilder will crash.

jhasse:
"Files changed" only shows this for me:

```diff
-		if(childVals.Count() > 2)
-		{
-			Item(2)->SetValue( childVals[2]);
-		}
+		//if(childVals.Count() > 2)
+		//{
+		//	Item(2)->SetValue( childVals[2]);
+		//}
```

I don't understand how this fixes the bug you described?!?

tomay3000:
it actually fixes the bug.
The changes I have made, just is commenting the code that trigers the bug.

Have you tried to reproduce it as I decribed (with and without the fix) ?

Thank you foryour understanding.

tomay3000:
Take a look at this screenshot to understand what I'm talking about:
untitled

jhasse:
While it does fix your bug (or workaround it), I'm not sure if it's okay to remove that. It might be used by other code in wxFormBuilder.

tomay3000:
I spent more than three days trying to understand Most of wxFormBuilder's code (believe me it is a mess!) .
if only the GUI was written using wxFormBuilder itself (it could be easier).

That bug I am talking about, will make wxFormBuilder crash also if you try to edit these property values.
if only anybody else could reproduce it and then apply the patch to see the results, then it will confirmed.

jhasse:
> believe me it is a mess!

I know ;)

> That bug I am talking about, will make wxFormBuilder crash also if you try to edit these property values.

I couldn't reproduce the crash, only the non-sense values. If I understand it correctly your patch simply doesn't update the ico_size property, which avoids the garbage values, but still doesn't sound like the final solution for me.

Also `wxFBBitmapProperty::UpdateChildValues` might be used by other parts which now won't work? I really don't know, unfortunately there are no automated tests for any of this :(

ctfyui:
@tomay3000 It seems we have the same problem (#419).

tomay3000:
@ctfyui I did fixed it in my wxFormBuilder version, take a look at the changed files.

jhasse:
Closing in favor of #425

tomay3000:
Thank you.

PR #389, opened 2018-05-26
Add support of markup for StaticText

eldarkg:


output/plugins/common/xml/common.cppcode
```diff
@@ -86,8 +86,9 @@ Written by
   <templates class="wxStaticText">
     <template name="declaration">#class* $name;</template>
     <template name="include">@#include &lt;wx/stattext.h&gt;</template>
-    <template name="construction">$name = new #class( #wxparent $name, $id, $label, $pos, $size, $style #ifnotnull $window_style @{ |$window_style @} #ifnotnull $window_name @{, $window_name @} );</template>
-  	<template name="settings">$name->Wrap( $wrap );</template>
+    <template name="construction">$name = new #class( #wxparent $name, $id, wxEmptyString, $pos, $size, $style #ifnotnull $window_style @{ |$window_style @} #ifnotnull $window_name @{, $window_name @} );</template>
```
jhasse: To reduce the code needed when not using markup, I would still pass the label to the constructor here, when `$markup` doesn't equal 1. Ideally the output shouldn't change at all when updating wxFormBuilder without changing a setting.

eldarkg: OK

plugins/common/common.cpp
```diff
@@ -368,14 +368,20 @@ class StaticTextComponent : public ComponentBase
...
-			obj->GetPropertyAsString(_("label")),
-			obj->GetPropertyAsPoint(_("pos")),
-			obj->GetPropertyAsSize(_("size")),
-			obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
+		wxString label = obj->GetPropertyAsString( _("label") );
+		wxStaticText* st = new wxStaticText( (wxWindow *)parent, -1,
+			label,
+			obj->GetPropertyAsPoint( _("pos") ),
+			obj->GetPropertyAsSize( _("size") ),
+			obj->GetPropertyAsInteger( _("style") ) | obj->GetPropertyAsInteger( _("window_style") ) );
```
jhasse: Please don't include the added spaces as it destroys the diff history. Sorry for the trouble and that I didn't mention this earlier :/

jhasse:
Thanks :)

PR #390, opened 2018-05-26
 Add support of markup for Button

eldarkg:

PR #417, opened 2018-07-06
Update codegeneration for wxWidgets 3.1, fix #211

### LINKED ISSUE #211: Codegeneration outdated for wxWidgets 3.1
The codegeneration templates and the xml files describing the object properties are out there since wxWidgets 2.8 and most of them dont have been updated for quite some time. wxWidgets is currently at version 3.1, wxFormBuilder itself cannot be compiled with wxWidgets 2.8 anymore, so i have taken the time to review ALL templates and xml files and updated them to wxWidgets 3.1 level. I have taken wxWidgets svn revision 76938 as source.

What i have done in detail:
- Reviewed all style flags, removed old ones, added new ones, renamed from old names to new names
- Reviewed all event types, added new event handlers, removed old ones. I did not add support for the new bind system, i have still to evaluate how its new features can be exposed to wxFormBuilder.
- Updated almost all help messages related to style flags and event handlers to the wording of the wxWidgets documentation
- I did not add any new controls or checked if additional parameters should be added to existing ones except i updated wxButton to support bitmaps now, wxButton is pretty much identical to wxBitmapButton now, but i kept wxBitmapButton for backward compatibility
- Updated font parameters (family, style, weight) to use the new names, this way generated code doesn't cause deprecated warnings anymore. I also updated the wxFormBuilder code itself to use the new names so compilation of wxFormBuilder also doesn't cause these warnings anymore.
- All these changes are implemented for ALL supported generation languages, including XRC, and of course the preview inside wxFormBuilder
- Updated project file version to 1.14 and implemented upgrading from the previous version

Because the codegeneration templates and the xml files had mixed whitespace and non uniform identation and because i had to edit them i took the chance to reformat every template and xml file. I have used the most commonly used style for each, that is tab inside the templates and space inside the xml files. This causes a lot of differences in the patch file but it was just too much work to seperate the whitespace changes from the functional changes.

All these changes are contained in the attached file codegen_wx31.patch.

Additionally i changed the codegeneration to use the static box of a wxStaticBoxSizer as parent for its children. However, like with the highlight patch for the designer, this was not easy and might have some side effects, therefor i made an extra patch for this, codegen_wx31_sbox.patch, this patch does not rely on the other one.

Basically the #wxparent search function now stops at a wxStaticBoxSizer and returns it, the following code generates the method call to return its static box. Usually the #wxparent macro is just used to get the name of the parent to use it as parent for the created element, but there is another case when it can be used, to extract parameters of it. This seems to be used only for the #ifnull and the like macros and it might be an issue that these return the wxStaticBoxSizer object now that isnt even a wxWindow. However the wxStaticBox is not part of the wxFormBuilder object hierarchy so there cannot be done much else. I analysed all generator templates and none seems to use #wxparent in a way that causes such a problem.

I am also not familiar with other generation languages than c++ so the method call to get the static box of the sizer might be wrong for these, but it looks pretty much like other code that gets generated for these languages :).

Reported by: sodev

jhasse:
This is the patch by @sodevel submitted here: https://github.com/wxFormBuilder/wxFormBuilder/issues/211#issuecomment-402268318

I've tried to get rid of all the formatting changes.

@sodevel: Do you have an email address that you would be okay to share? If so, I would change the author of the commit to you.


output/plugins/layout/xml/layout.xml
```diff
@@ -76,12 +76,20 @@ Written by
...
 
-  <objectinfo class="wxWrapSizer" icon="sizer.xpm" type="sizer">
+  <objectinfo class="wxWrapSizer" icon="wrap_sizer.xpm" type="sizer">
 	<inherits class="sizer" />
     <property name="name" type="text" hidden="1">wSizer</property>
     <property name="orient" type="option">
       <option name="wxVERTICAL" 	help="Align items vertically"/>
-      <option name="wxHORIZONTAL" 	help="Align items horizontally"/>wxHORIZONTAL</property>
+      <option name="wxHORIZONTAL" 	help="Align items horizontally"/>
+      wxVERTICAL
```
jhasse: If I understand you correctly [here](https://github.com/wxFormBuilder/wxFormBuilder/issues/211#issuecomment-402530838), it would be better to change this to `wxHORIZONTAL`, right?

sodevel: Yes, because that is the default value of this sizer

jhasse:
I've updated the author and changed the default to `wxHORIZONTAL`.

src/rad/appdata.cpp
```diff
@@ -1953,7 +1953,192 @@ void ApplicationData::ConvertObject( ticpp::Element* parent, int fileMajor, int
...
+				if (TypeConv::FlagSet(wxT("wxNO_BORDER"), styles)) {
+					styles = TypeConv::ClearFlag(wxT("wxNO_BORDER"), styles);
+					styles = TypeConv::SetFlag(wxT("wxBORDER_NONE"), styles);
+				}
+
+				(*newProp)->SetText(_STDSTR(styles));
+			}
+		}
+
+		// wxBitmapButton: Remove wxBU_AUTODRAW and rename properties selected->pressed,
```
jhasse: I've found that `wxBU_AUTODRAW` is marked obsolete in wxWidgets' source. Is there any documentation on this?

Why does the `wxBitmapButton` constructor still pass it as a style? http://docs.wxwidgets.org/3.1/classwx_bitmap_button.html

output/plugins/common/xml/common.cppcode
```diff
@@ -33,6 +33,27 @@ Written by
 	    #ifnotequal $default "0" @{
 		    $name->SetDefault(); #nl
 	    @}
+
+			#ifnotnull $bitmap
+			@{ #nl $name->SetBitmap( $bitmap ); @}
```
jhasse: Can't this be removed as `$bitmap` is already passed to the constructor?

src/rad/appdata.cpp
sodevel: ``wxBitmapButton`` itself is obsolete. I think that style got simply removed from the docs when they did their ``wxAnyButton`` refactoring.

output/plugins/common/xml/common.cppcode
sodevel: This hunk is inside wxButton, that doesn't get the bitmap from the constructor

jhasse: It still gets unnecessarily added for `wxBitmapButton` (because it inherits from `wxButton`?) though.

sodevel: Interesting, its even worse than that. The settings templates of all inherited classes get applied and because the bitmaps get set in both this results in setting all the other bitmaps twice. Nice catch, i didn't discover this in all these years i was using that version.

I'm going to fix that, plus im going to fix #390, the markup isn't stored in XRC.

PR #418, opened 2018-07-06
Improve buildsystem

sodevel:
This removes elements for building for wxWidgets < 2.9, many of them referenced non existing elements anyway. Additionally it improves building with MSVC by not passing GCC specific parameters anymore and disabling secure CRT warnings with newer macros and in more locations.


jhasse:
Can you squash ea14d01 and 0a7871b? To do that run

`git rebase -i`

which will open an editor. There you can change `pick` to `squash` for 0a7871b.

Also please don't start the first commit message line with `- `.

jhasse:
Thanks!

Btw: The long term goal is to replace the Premake build system with CMake. Work has stalled on that though, but I'll commit Meson build files soon which I've been working on.

PR #423, opened 2018-07-11
Fix warnings

sodevel:
Fixes various warnings reported by MSVC 2015 except wxWidgets deprecation warnings which are handled by #417. Bigger rework of the MD5 class to prevent implicit type conversions.

### DIFF
--- a/src/md5/md5.cc
+++ b/src/md5/md5.cc
@@ -120,11 +120,11 @@ void MD5::update(const uint8_t* input, uint32_t input_length) {
 
 void MD5::update(FILE *file){
 
-  unsigned char buffer[1024];
-  int len;
+  unsigned char fileBuffer[1024];
+  size_t len;
 
-  while ((len=fread(buffer, 1, 1024, file)))
-    update(buffer, len);
+  while ((len=fread(fileBuffer, 1, 1024, file)))
+    update(fileBuffer, len);
 
   fclose (file);
 
@@ -140,13 +140,13 @@ void MD5::update(FILE *file){
 
 void MD5::update(istream& stream){
 
-  unsigned char buffer[1024];
-  int len;
+  unsigned char streamBuffer[1024];
+  std::streamsize len;
 
   while (stream.good()){
-    stream.read((char*)buffer, 1024); // note that return value of read is unusable.
+    stream.read((char*)streamBuffer, 1024); // note that return value of read is unusable.
     len=stream.gcount();
-    update(buffer, len);
+    update(streamBuffer, static_cast<unsigned int>(len));
   }
 
 }
@@ -161,13 +161,13 @@ void MD5::update(istream& stream){
 
 void MD5::update(ifstream& stream){
 
-  unsigned char buffer[1024];
-  int len;
+  unsigned char streamBuffer[1024];
+  std::streamsize len;
 
   while (stream.good()){
-    stream.read((char*)buffer, 1024); // note that return value of read is unusable.
+    stream.read((char*)streamBuffer, 1024); // note that return value of read is unusable.
     len=stream.gcount();
-    update(buffer, len);
+    update(streamBuffer, static_cast<unsigned int>(len));
   }
 
 }
--- a/src/md5/md5.cc
+++ b/src/md5/md5.cc
@@ -47,11 +47,8 @@ documentation and/or software.
 
 #include "md5.hh"
 
-#include <assert.h>
+#include <cassert>
 #include <cstring>
-#include <string>
-
-using namespace std;
 
 
 // MD5 simple initialization method
@@ -118,7 +115,7 @@ void MD5::update(FILE* file) {
 
 	assert(!finalized);
 
-	while ((len = fread(fileBuffer, 1, 1024, file)))
+	while ((len = std::fread(fileBuffer, 1, 1024, file)))
 		update(fileBuffer, static_cast<uint32_t>(len));
 }
 
@@ -130,7 +127,7 @@ void MD5::update(FILE* file) {
 // MD5 update for istreams.
 // Like update for files; see above.
 
-void MD5::update(istream& stream) {
+void MD5::update(std::istream& stream) {
 	char streamBuffer[1024];
 	std::streamsize len;
 
@@ -206,7 +203,7 @@ MD5::MD5(FILE* file) {
 
 
 
-MD5::MD5(istream& stream) {
+MD5::MD5(std::istream& stream) {
 	init(); // must be called by all constructors
 	update(stream);
 	finalize();
--- a/src/md5/md5.cc
+++ b/src/md5/md5.cc
@@ -50,7 +50,6 @@ documentation and/or software.
 #include <assert.h>
 #include <cstring>
 #include <string>
-#include <iostream>
 
 using namespace std;
 
@@ -75,10 +74,7 @@ void MD5::update(const unsigned char* input, uint32_t input_length) {
 	uint32_t input_index, buffer_index;
 	uint32_t buffer_space; // how much space is left in buffer
 
-  if (finalized){  // so we can't update!
-    cerr << "MD5::update:  Can't update a finalized digest!" << endl;
-    return;
-  }
+	assert(!finalized);
 
   // Compute number of bytes mod 64
   buffer_index = (count[0] >> 3) & 0x3F;
@@ -123,11 +119,10 @@ void MD5::update(FILE *file){
   unsigned char fileBuffer[1024];
   size_t len;
 
+	assert(!finalized);
+
   while ((len=fread(fileBuffer, 1, 1024, file)))
     update(fileBuffer, static_cast<uint32_t>(len));
-
-  fclose (file);
-
 }
 
 
@@ -143,7 +138,9 @@ void MD5::update(istream& stream){
   char streamBuffer[1024];
   std::streamsize len;
 
-  while (stream.good()){
+	assert(!finalized);
+
+  while (stream) {
     stream.read(streamBuffer, 1024); // note that return value of read is unusable.
     len=stream.gcount();
     update(reinterpret_cast<unsigned char*>(streamBuffer), static_cast<uint32_t>(len));
@@ -170,10 +167,7 @@ void MD5::finalize (){
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
     };
 
-  if (finalized){
-    cerr << "MD5::finalize:  Already finalized this digest!" << endl;
-    return;
-  }
+	assert(!finalized);
 
   // Save number of bits
   encode (bits, count, 8);
@@ -231,12 +225,7 @@ unsigned char *MD5::raw_digest(){
 
   unsigned char *s = new unsigned char[16];
 
-  if (!finalized){
-    cerr << "MD5::raw_digest:  Can't get digest if you haven't "<<
-      "finalized the digest!" <<endl;
-		std::memset(s, 0, 16);
-		return s;
-  }
+	assert(finalized);
 
 	std::memcpy(s, digest, 16);
   return s;
@@ -247,12 +236,7 @@ unsigned char *MD5::raw_digest(){
 char *MD5::hex_digest(){
   char *s= new char[33];
 
-  if (!finalized){
-    cerr << "MD5::hex_digest:  Can't get digest if you haven't "<<
-      "finalized the digest!" <<endl;
-	s[0]='\0';
-    return s;
-  }
+	assert(finalized);
 
   for (int i=0; i<16; i++)
     sprintf(s+i*2, "%02x", digest[i]);
--- a/src/md5/md5.cc
+++ b/src/md5/md5.cc
@@ -57,9 +57,7 @@ using namespace std;
 // MD5 simple initialization method
 
 MD5::MD5(){
-
-  init();
-
+	init();
 }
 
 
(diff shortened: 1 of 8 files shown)


plugins/additional/additional.cpp
```diff
@@ -1320,12 +1320,12 @@ class SearchCtrlComponent : public ComponentBase
 
 		if ( !obj->IsNull( _("search_button") ) )
 		{
-			sc->ShowSearchButton( obj->GetPropertyAsInteger( _("search_button") ) );
+			sc->ShowSearchButton( !!obj->GetPropertyAsInteger( _("search_button") ) );
```
jhasse: I think `!= 0` would be clearer.

plugins/additional/additional.cpp
```diff
@@ -1806,24 +1806,24 @@ void ComponentEvtHandler::OnMarginClick( wxStyledTextEvent& event )
 class DataViewModel : public wxDataViewModel
 {
 public:
-	unsigned int GetChildren(const wxDataViewItem& item, wxDataViewItemArray& children) const {
+	unsigned int GetChildren(const wxDataViewItem& WXUNUSED(item), wxDataViewItemArray& WXUNUSED(children)) const {
```
jhasse: Are there any advantages of `WXUNUSED` over just commenting out the name with `/* */`?

src/md5/md5.cc
```diff
@@ -80,7 +80,7 @@ void MD5::update ( const uint1 *input, uint4 input_length) {
   }
 
   // Compute number of bytes mod 64
-  buffer_index = (unsigned int)((count[0] >> 3) & 0x3F);
+  buffer_index = (uint4)((count[0] >> 3) & 0x3F);
```
jhasse: `uint32_t` would be even better. Why change the type at all though?

src/md5/md5.cc
```diff
@@ -119,11 +119,11 @@ void MD5::update ( const uint1 *input, uint4 input_length) {
 
 void MD5::update(FILE *file){
 
-  unsigned char buffer[1024];
-  int len;
+  unsigned char fileBuffer[1024];
+  size_t len;
```
jhasse: Why not `uint4`/`uint32_t`? Then you wouldn't need the `static_cast`.

src/md5/md5.cc
```diff
@@ -119,11 +119,11 @@ void MD5::update ( const uint1 *input, uint4 input_length) {
 
 void MD5::update(FILE *file){
 
-  unsigned char buffer[1024];
-  int len;
+  unsigned char fileBuffer[1024];
```
jhasse: Any reason for the rename?

src/md5/md5.cc
```diff
@@ -219,6 +198,15 @@ void MD5::finalize (){
 
 
 
+MD5::MD5(const unsigned char* input, uint4 input_length) {
+	init();  // must be called be all constructors
```
jhasse: Typo: by

src/rad/wxfbmanager.cpp
```diff
@@ -155,16 +155,16 @@ wxObject* wxFBManager::GetWxObject( PObjectBase obj )
 
 void wxFBManager::ModifyProperty( wxObject* wxobject, wxString property, wxString value, bool allowUndo )
 {
-	CHECK_VISUAL_EDITOR()
+	CHECK_VISUAL_EDITOR(void())
```
jhasse: Why is this needed?

plugins/additional/additional.cpp
sodevel: You can debate if the double bang is clearer, but it is shorter and more flexible to normalize X to a bool. It can also be used in cases where X can't be compared to an int, e.g. the safe bool idiom or a ``wxStreamBase`` that doesn't convert to a bool at all and only has ``operator!``.

plugins/additional/additional.cpp
sodevel: Portability. I don't know how much this is an issue today, but i remember having read that some compilers don't support this and require other means to ignore unused parameters. Thats why WXUNUSED, boost::ignore_unused() and the like exist, so i used the wxWidgets style here.

src/md5/md5.cc
sodevel: The type usage in this class was a huge mess, i tried the minimum change approach to fix it. ``buffer_index`` is a ``uint4`` so i cast the right side to the left side. Most probably this cast isn't necessary at all because ``count`` is a ``uint4`` as well and the integer literals are so small that they won't introduce a type cast at all, but the cast was there so i kept it.

src/md5/md5.cc
sodevel: Because ``fread`` returns a ``size_t`` which may be one of these but also can be ``uint64_t`` or whatever ``typedef /*implementation-defined*/ size_t;`` is in the used implementation. So in these locations i used the types native to the functions that write them and cast them to the types the functions need that read them.

src/md5/md5.cc
sodevel: To get rid of the shadow warnings caused by the class member with the same name. This is the reason for the various other renames as well.

src/md5/md5.cc
sodevel: I shouldn't have copy-pasted the file-constructor but the stream-constructor, there its correct...

src/rad/wxfbmanager.cpp
sodevel: This macro requires a parameter and MSVC spills out a warning if you don't supply one. There a tricks to produce macros that accept a variable amount of parameters or you can make two macros, one for methods that return something and one for methods that return void (wxWidgets uses this for its assert macros), but this was the least-amount-of-work solution.

jhasse: Can you disable the warning for MSVC instead?

plugins/additional/additional.cpp
jhasse: Okay, I think that isn't an issue today anymore.

plugins/additional/additional.cpp
jhasse: In this case it actually *is* an integer though.

src/rad/wxfbmanager.cpp
sodevel: Interesting, according to the standard it is even an error to supply less arguments than required, but empty arguments are ok. For some reason [gcc](https://gcc.gnu.org/onlinedocs/cpp/Macro-Arguments.html) interprets empty parentheses as one empty argument but MSVC not, that one however doesn't produce an error if you don't supply enough.

Regarding that these macros are only used local in that file i guess i can disable the warning instead.

plugins/additional/additional.cpp
sodevel: It is/was wxWidgets style to suppress these warnings, i would keep them so its easier in the future to replace them with [[maybe_unused]]

plugins/additional/additional.cpp
sodevel: Still it is shorter and has its semantic meaning of "normalize-to-bool-operator". I am used to use it in my codebase, so what shall i do?

sodevel:
I pushed a replacement for df72141 but kept it. I can rebase this PR to current master to fix that trivial merge conflict and remove that commit and do other cleanup like maybe splitting the MD5 refactoring which is a little more than just a warning fix but for now i keep it as is to prevent issues with github.

plugins/additional/additional.cpp
jhasse: `[[maybe_unused]]` is even longer and "maybe" just seems like "disable the warning" to me. But it's a good warning because it can catch some problems (e.g. mistyping) or unnecessary parameters.

In this case, I think it's actually best to just remove the parameter name, as "item" doesn't add any extra information anyway.

I've fixed the warning with `/* */` or by removing the name in 0985941f5e4dac3c1a46ae029cbb66cd257d1b97.

src/rad/wxfbmanager.cpp
```diff
@@ -26,6 +26,12 @@
 #include "model/objectbase.h"
 #include "rad/appdata.h"
 
+// Sometimes these macros get called with an empty argument which is not an error
+// but MSVC still produces a warning about too few arguments, disable it
+#ifdef _MSC_VER
+#pragma warning(disable:4003)
```
jhasse: You can also disable warnings in MSVC's project files, which I would prefer, as I don't like to add MSVC specific code in cpp files.

(I actually don't like to support MSVC at all, as it increases the maintenance burden and is proprietary)

jhasse:
I've cherry-picked a few commits from this PR to the master branch (messed it up actually, so CI failed. Sorry about that!).

It might be a little bit harder for you to rebase this now. Sorry again ...

src/rad/wxfbmanager.cpp
sodevel: Since MSVC is like the "native" compiler on windows i am quite interested that it is supported. Plus compiling wxWidgets on my machine with MSVC running on one core is still faster than with gcc running on all four.

sodevel:
Rebased on current master

src/md5/md5.cc
```diff
@@ -215,31 +199,32 @@ void MD5::finalize (){
...
 
 
 
-MD5::MD5(FILE *file){
-
-	init(); // must be called by all constructors
-  update(file);
-  finalize ();
+MD5::MD5(const unsigned char* input, uint32_t input_length) {
+	init();  // must be called be all constructors
```
jhasse: Introduces the typo again.

src/md5/md5.cc
```diff
@@ -215,31 +199,32 @@ void MD5::finalize (){
...
 
 
-MD5::MD5(istream& stream){
+MD5::MD5(FILE *file){
 
 	init(); // must be called by all constructors
-  update (stream);
-  finalize();
+  update(file);
+  finalize ();
```
jhasse: Please remove the space.

src/md5/md5.cc
```diff
@@ -250,11 +235,9 @@ MD5::MD5(ifstream& stream){
 
 unsigned char *MD5::raw_digest(){
 
-	uint8_t* s = new uint8_t[16];
+  unsigned char *s = new unsigned char[16];
 
   if (!finalized){
-    cerr << "MD5::raw_digest:  Can't get digest if you haven't "<<
-      "finalized the digest!" <<endl;
```
jhasse: Aren't those error messages a good thing? Now the caller doesn't have an idea why it failed.

plugins/additional/additional.cpp
jhasse: If it isn't okay to implicitly cast an int to bool, I don't think that passing an int to `!` should be allowed either.

> safe bool idiom

If I understand it correctly this is obsolete now that C++11 supports explicit conversion operators.

build/premake/wxformbuilder.lua
```diff
@@ -72,6 +72,7 @@ project "wxFormBuilder"
 
     configuration "vs*"
         defines         {"_CRT_SECURE_NO_DEPRECATE", "_CRT_SECURE_NO_WARNINGS"}
+        buildoptions    {"/wd4003"}
```
jhasse: Perfect, thanks :)

src/md5/md5.cc
sodevel: I had so much fun with interactive rebasing and editing, squashing, splitting commits, i guess some fun slipped through the cage, will fix this later

src/md5/md5.cc
sodevel: This is sort of a library and it shouldn't produce unexpected console output. Plus this doesn't work with a windows GUI application anyway. Plus this doesn't help you to catch this error programmatically.

On second thought, this isn't even a real error condition but a programmer error that does only deserve an assertion, like some code below already does. It's a much more serious error to actually ignore the read errors from the FILE/stream.

It's just amazing how many issues are in this tiny piece of code, i just wanted to fix some warnings...

plugins/additional/additional.cpp
sodevel: This doesn't stop [MSVC](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4800) to complain about

sodevel: I see you don't like my double bang operator, i could go the /wd4800 route if you prefer

jhasse: > This doesn't stop MSVC to complain about

What do you mean by "this"?

> I see you don't like my double bang operator, i could go the /wd4800 route if you prefer

I still think `!= 0` would be a good idea, it will also silence MSVC's warning. I will push something tomorrow.

src/md5/md5.cc
jhasse: I see. What do you think about throwing a `std::runtime_error`?

plugins/additional/additional.cpp
sodevel: > What do you mean by "this"?

Explicit conversion

src/md5/md5.cc
sodevel: The exception for what? Getting called in the wrong finalization state? Or the read errors? I wouldn't use an exception here for both cases, this whole "library" is exception free and looks so much c-like and the calling side is not prepared for this, its not worth the trouble. Assert for the finalization and if necessary use a return code for read errors.

jhasse: The calling side isn't prepared for an assert either though. And what "trouble" do you mean?

sodevel: Maybe there is a misunderstanding what an assert is, so lets clear this up: In release mode an assert does nothing, in debug mode, if it triggers, it displays the message and then terminates the program, so there is no calling side anymore ;).

wxWidgets has these nice "hybrid-asserts", they are also active in release mode and silently perform the trigger-action there, in debug mode they pop up the message and let you choose what to do. These are a "nicer" version of a "true" assert, they allow the program to continue with some meaningful value in cause of the assert.

The "trouble" with exceptions here is that they are not so common like in java and here there is also no such thing like checked exceptions. So you have to write some documentation like "This does throw this and that exception in case of error". The TiCPP library does this, wxFB itself uses internally its own exception class, but this class is so small, does nasty c-style things like returning something you need to delete, basically the only c++-element is the istream (ok, and ostream inserter), an error code would fit here much better in my opinion.

With some small modifications you don't even need to signal read errors at all, you could let the caller check the error state of the stream to detect an error (currently it always will be in error state because of the way it is read from, but this can be made better). For the FILE version you need to stop to delete it in the method and it would work there as well. That would give a nice interface that is identical to the memory block version.

jhasse:
While some changes could be squashed into fewer commits, I think that's nitpicking at this point. Thank you very much for the changes, highly appreciated!

Issue #421, opened 2018-07-09
Unique Includes in LuaCodeGenerator broken


sodevel:
While fixing warnings i discovered this in `src/codegen/luacg.cpp`:

```
void LuaCodeGenerator::AddUniqueIncludes( const wxString& include, std::vector< wxString >* includes )
{
	// Split on newlines to only generate unique include lines
	// This strips blank lines and trims
	wxStringTokenizer tkz( include, wxT("\n"), wxTOKEN_STRTOK );

	while ( tkz.HasMoreTokens() )
	{
		wxString line = tkz.GetNextToken();
		line.Trim( false );
		line.Trim( true );


			includes->push_back( line );
			continue;

		// If it is an include, it must be unique to be written
		std::vector< wxString >::iterator it = std::find( includes->begin(), includes->end(), line );
		if ( includes->end() == it )
		{
			includes->push_back( line );
		}
	}
}
```

Apparently that unique check never gets executed. I think the lines
```
			includes->push_back( line );
			continue;
```
don't belong there and should be removed, but im not familiar enough especially with LUA so i'm not sure.

jhasse:
I also don't know Lua, but I think you're right.

jhasse:
Should be fixed by the PR.

PR #424, opened 2018-07-11
Fixes LUA unique include filtering

sodevel:
This fixes the issue #421. The solution was found by looking at the generated code, for some reason unique filtering accross different objects was working, only the filtering in a single object was affected.


jhasse:
Thanks!

Minor nitpick: Please use imperative mood in the title of the commit message, e.g. "Fix Lua unique ..."

Issue #419, opened 2018-07-07
Bug: Program crash when modifying icon size in wxStaticBitmap


ctfyui:
Program crash when modifying iconsize in wxStaticBitmap

![Screenshot](https://user-images.githubusercontent.com/33893082/42407905-6ae0a872-81f6-11e8-8375-6e3af2511dce.png)

Error infomation

14:57:32: 0x75bf98cd: free in C:\Windows\syswow64\msvcrt.dll
14:57:32: 0x7151ada9: ZN19wxVariantDataStringD0Ev in D:\Program Files (x86)\wxFormBuilder\wxbase30u_gcc_custom.dll
14:57:32: 0x70d37b36: ZN14wxPropertyGrid17DoPropertyChangedEP12wxPGPropertyj in D:\Program Files (x86)\wxFormBuilder\wxmsw30u_propgrid_gcc_custom.dll
14:57:32: 0x70d32ab0: ZN14wxPropertyGrid23HandleCustomEditorEventER7wxEvent in D:\Program Files (x86)\wxFormBuilder\wxmsw30u_propgrid_gcc_custom.dll
14:57:32: 0x70d69f4a: ZN34wxPropertyGridEditorEventForwarder12ProcessEventER7wxEvent in D:\Program Files (x86)\wxFormBuilder\wxmsw30u_propgrid_gcc_custom.dll
14:57:32: 0x6261f626: ZN10wxTextCtrl6OnCharER10wxKeyEvent in D:\Program Files (x86)\wxFormBuilder\wxmsw30u_core_gcc_custom.dll
14:57:32: 0x71382bc2: ZNK16wxAppConsoleBase11HandleEventEP12wxEvtHandlerMS0_FvR7wxEventES3_ in D:\Program Files (x86)\wxFormBuilder\wxbase30u_gcc_custom.dll
14:57:32: 0x75bf98da: free in C:\Windows\syswow64\msvcrt.dll
14:57:32: 0x71464ef7: ZN9wxVariantD1Ev in D:\Program Files (x86)\wxFormBuilder\wxbase30u_gcc_custom.dll
14:57:32: 0x70d0cf07: ZNK18wxPGTextCtrlEditor7OnEventEP14wxPropertyGridP12wxPGPropertyP8wxWindowR7wxEvent in D:\Program Files (x86)\wxFormBuilder\wxmsw30u_propgrid_gcc_custom.dll
[...]

PR #425, opened 2018-07-12
Fix garbage ico_size values

sodevel:
This fixes the root cause of #419 and #376. The reported crash probably happened because the preview tried to create an icon with these too big garbage values.


jhasse:
Awesome! Thank you very much.

tomay3000:
Thank you.

Issue #427, opened 2018-07-13
Broken codegen for wxBitmapButton on macOS since wxFB 3.8


Zenju:
Hi,

This is in reference to https://github.com/wxFormBuilder/wxFormBuilder/pull/417

Since wxFB 3.8, the wxBitmapbuttons are drawn incorrectly on macOS:
![2](https://user-images.githubusercontent.com/7656530/42710622-fb82971c-86e4-11e8-9795-1ea5fddeda05.png)

This should look like:
![1](https://user-images.githubusercontent.com/7656530/42710621-fb619fc6-86e4-11e8-92ab-b62ff1a7e65a.png)

The reason is the lack of the "wxBU_AUTODRAW" style flag on wxBitmapButton. Even though it is (obviously incorrectly) marked as obsolete by wxWidgets themselves, it is still being used on macOS, even in the latest wxWidgets trunk code:
https://github.com/wxWidgets/wxWidgets/blob/0e92eb13dc127fd02e75f40ff83c1cc55ce4bdb4/src/osx/bmpbuttn_osx.cpp#L41

Please bring it back or at least provide a style flag that can be set in wxFB. Currently I don't see a workaround other than to revert back to the previous wxFB release.

Thanks for your awesome work!
Best

sodevel:
What happens when you use a regular `wxButton` with a bitmap and the styles `wxBU_EXACTFIT` and `wxBU_NOTEXT` instead of the `wxBitmapButton`?

Zenju:
wxButton + wxBU_NOTEXT + wxBU_EXACTFIT might be a viable alternative. I'll have to do some more testing, but the first attempt seems to work just like wxBitmapbutton on both macOS and Windows.
I'm however seeing one immediate issue:

The wxFB preview used to look like:
![1](https://user-images.githubusercontent.com/7656530/42715086-9f73f252-86f5-11e8-8112-9d0c2ed87a1d.png)

but with wxBitmapbuttons replaced by "wxButton + wxBU_NOTEXT + wxBU_EXACTFIT" looks like the following which is pretty much unusable at a large project-scope:
![2](https://user-images.githubusercontent.com/7656530/42715085-9f568bae-86f5-11e8-9332-1412902d7164.png)

sodevel:
I don't observe this effect here, looks the same for me in the preview. Are both of your versions really the same, no different loading type, size specifications or the like?

Zenju:
[...]

PR #428, opened 2018-07-14
wxButton fixups

sodevel:
Fixes the issues reported here #427 and here #417

### DIFF
--- a/output/plugins/common/xml/common.cppcode
+++ b/output/plugins/common/xml/common.cppcode
@@ -62,27 +62,8 @@ Written by
 
   <templates class="wxBitmapButton">
     <template name="declaration">#class* $name;</template>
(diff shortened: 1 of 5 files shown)


sodevel:
Maybe resurrecting wxBU_AUTODRAW is not a good idea, after all the project file upgrade just removed it so everyone has to add it again. I checked the wxWidgets sources and it seems osx is the only affected port, gtk and msw just create a normal wxButton with wxBU_EXACTFIT and wxBU_NOTEXT in the wxBitmapButton constructor, only osx adds extra margins when that style is passed. It looks like osx actually can not produce a correct result without that style, so maybe we should do what wxWidgets does by default, always add that style in the generator templates.

Looks like a regression in wxWidgets to me, should be further analyzed.

jhasse:
> after all the project file upgrade just removed it so everyone has to add it again.

We could still save people some trouble which haven't upgraded to 3.8.0 yet.

sodevel:
Since this is deprecated and most probably a bug in wxWidgets i would prefer to do the same like they do. They use it as default and most users probably won't specify style flags at all because there is no need so this is in fact always specified, simply specifiy this flags always in the generator templates for now.

jhasse:
Thanks for the PR.

PR #432, opened 2018-07-16
Adds the feature to load bitmap resources from XRC

sodevel:
A new option gets added to the wxFBBitmapProperty to load the bitmap from the XRC system. This allows using [Binary Resource Files](http://docs.wxwidgets.org/trunk/overview_xrc.html). It is up to the application code to initialize the resources. Currently this is only possible for the C++ target because of lack of knowledge if and how the other targets support this.

All other languages return a wxNullBitmap if this method is selected, this is what they already do if one of the other unsupported methods for the language is selected.

This is the second feature of my not-so-private-anymore fork of wxFormBuilder that got rejected in the past because it was only implemented for C++, maybe someone else can verify that the other languages don't support this or can implement this for them if possible.

### DIFF
--- a/src/rad/inspector/wxfbadvprops.cpp
+++ b/src/rad/inspector/wxfbadvprops.cpp
@@ -499,99 +499,106 @@ wxFBBitmapProperty::~wxFBBitmapProperty()
 }
 
 wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int childIndex,
-                                           wxVariant& childValue) const {
-	wxFBBitmapProperty* bp = (wxFBBitmapProperty*)this;
+                                           wxVariant& childValue) const
+{
+	auto* bp = const_cast<wxFBBitmapProperty*>(this);
 
-    wxString val = thisValue.GetString();
+	const auto val = thisValue.GetString();
 	wxArrayString childVals;
-	GetChildValues( val, childVals );
-	wxString newVal = val;
+	GetChildValues(val, childVals);
+	auto newVal = val;
 
-    // Find the appropriate new state
-    switch ( childIndex )
-    {
-        // source
-        case 0:
-        {
-			unsigned int count = GetChildCount();
+	// Find the appropriate new state
+	switch (childIndex)
+	{
+		// source
+		case 0:
+		{
+			const auto count = GetChildCount();
 
-            // childValue.GetInteger() returns the chosen item index
-            switch ( childValue.GetInteger() )
-            {
-                // 'Load From File' and 'Load From Embedded File'
-                case 0:
-                case 1:
-                {
-					if( prevSrc != 0 && prevSrc!= 1 )
+			// childValue.GetInteger() returns the chosen item index
+			switch (childValue.GetInteger())
+			{
+				// 'Load From File' and 'Load From Embedded File'
+				case 0:
+				case 1:
+				{
+					if (prevSrc != 0 && prevSrc!= 1)
 					{
-					    for ( unsigned int i = 1; i<count ;i++ )
+						for (unsigned int i = 1; i < count ; ++i)
 						{
-							wxPGProperty *p = Item(i) ;
-							if ( p )
+							if (auto* p = Item(i))
 							{
-								wxLogDebug( wxT("wxFBBP::ChildChanged: Removing:%s"), p->GetLabel().c_str() );
-								GetGrid()->DeleteProperty( p );
+								wxLogDebug(wxT("wxFBBP::ChildChanged: Removing:%s"), p->GetLabel().c_str());
+								GetGrid()->DeleteProperty(p);
 							}
 						}
-						bp->AppendChild( bp->CreatePropertyFilePath() );
+						bp->AppendChild(bp->CreatePropertyFilePath());
 					}
 
-					if( childVals.GetCount() == 2 )
+					if (childVals.GetCount() == 2)
+					{
 						newVal = childVals.Item(0) + wxT("; ") + childVals.Item(1);
-					else if( childVals.GetCount() > 1 )
-						newVal = childVals.Item(0) + wxT("; ");
-
-                    break;
-                }
-                // 'Load From Resource'
-                case 2:
-                {
-					if( prevSrc != 2 )
+					}
+					else if (childVals.GetCount() > 0)
 					{
-						for ( unsigned int i = 1; i<count ;i++ )
+						newVal = childVals.Item(0) + wxT("; ");
+					}
+					break;
+				}
+				// 'Load From Resource'
+				case 2:
+				{
+					if (prevSrc != 2)
+					{
+						for (unsigned int i = 1; i < count ; ++i)
 						{
-							wxPGProperty *p = Item(i) ;
-							if ( p )
+							if (auto* p = Item(i))
 							{
-								wxLogDebug( wxT("wxFBBP::ChildChanged: Removing:%s"), p->GetLabel().c_str() );
-								GetGrid()->DeleteProperty( p );
+								wxLogDebug(wxT("wxFBBP::ChildChanged: Removing:%s"), p->GetLabel().c_str());
+								GetGrid()->DeleteProperty(p);
 							}
 						}
-						bp->AppendChild( bp->CreatePropertyResourceName() );
+						bp->AppendChild(bp->CreatePropertyResourceName());
 					}
 
-					if( childVals.GetCount() == 2)
+					if (childVals.GetCount() == 2)
+					{
 						newVal = childVals.Item(0) + wxT("; ") + childVals.Item(1);
-					else if( childVals.GetCount() > 1 )
-						newVal = childVals.Item(0) + wxT("; ");
-
-                    break;
-                }
-                // 'Load From Icon Resource'
-                case 3:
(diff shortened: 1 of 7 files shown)


src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -578,10 +594,33 @@ wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int child
...
+						newVal = childVals.Item(0) + wxT("; ") + childVals.Item(1);
+					else if (childVals.GetCount() > 1)
+						newVal = childVals.Item(0) + wxT("; ");
+					break;
+				}
+				// 'Load From Art Provider'
+                case 5:
                 {
-					if( prevSrc != 4 )
+					if( prevSrc != 5 )
```
jhasse: Hm ... we should do something about these magic numbers.

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -606,7 +645,7 @@ wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int child
             break;
         }
 
-        // file_path || id || resource_name
+        // file_path || id || resource_name || xrc_name
```
jhasse: What does this comment mean?

sodevel: This comment shows the name of the value the child at this index represents for the source types that use it (all source types currently use this child index). Hmm, this sounds quite complicated, this is the name of the row in the property grid below the row source 😄.

src/rad/inspector/wxfbadvprops.cpp
sodevel: This is one of the smaller problems, a much bigger problem is not only the usage of the display strings as data storage but also directly writing them into the project file. Currently its not possible to localize wxFB without creating localized project files that can only be processed with a wxFB using the same locale. Plus there are too many strings marked for translation, but this is not that hard to fix.

jhasse: I don't think wxFB will ever by localized (and IMHO it also never should be).

sodevel:
I have added implementations for Python and PHP.

Disclaimer: I don't know these languages nor their wxWidgets port, i've read some documenation and the generated code looks correct for me, but i don't know if it really is. Would be nice if someone could verify that.

The Lua part is more tricky, it does support XRC but there is no global wxXmlResource object. So the application code is required to create one in a manner compatible for the code generator. I don't know if it can be put into the UI table with a configured name so the generator can pick it up from there, maybe someone with more wxLua knowledge can help.

sodevel:
I've digged around more in the wxLua documentation and it seems you can set a global wxXmlResource object there as well, i have added a Lua implementation based on that.

This feature is now completed for all languauges.

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -578,10 +594,33 @@ wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int child
...
-                // 'Load From Art Provider'
-                case 4:
+				// 'Load From XRC'
+				case 4:
+				{
+					if (prevSrc != 4)
+					{
+						for (unsigned int i = 1; i<count; i++)
+						{
+							wxPGProperty *p = Item(i);
```
jhasse: Move this into the `if( )`

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -694,6 +733,13 @@ void wxFBBitmapProperty::UpdateChildValues(const wxString& value)
 			Item(2)->SetValue(WXVARIANT(TypeConv::StringToSize(aux)));
 		}
 	}
+	else if (childVals[0].Contains(_("Load From XRC")))
```
jhasse: here too

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -242,6 +246,7 @@ wxPGProperty *wxFBBitmapProperty::CreatePropertySource( int sourceIndex )
     sourceChoices.Add(_("Load From Embedded File") );
     sourceChoices.Add(_("Load From Resource") );
     sourceChoices.Add(_("Load From Icon Resource") );
+	sourceChoices.Add(_("Load From XRC"));
```
jhasse: here too

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -578,10 +594,33 @@ wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int child
...
+								GetGrid()->DeleteProperty(p);
+							}
+						}
+						bp->AppendChild(bp->CreatePropertyXrcName());
+					}
+
+					if (childVals.GetCount() == 2)
+						newVal = childVals.Item(0) + wxT("; ") + childVals.Item(1);
+					else if (childVals.GetCount() > 1)
+						newVal = childVals.Item(0) + wxT("; ");
```
jhasse: Hm ... why is `== 2` a special case?

sodevel: Because copy-paste. Each case uses this code fragment, some check for `== 2`, some for `== 3`, doesn't make sense for the `== 2` case, but looks uniform.

src/rad/inspector/wxfbadvprops.cpp
```diff
@@ -578,10 +594,33 @@ wxVariant wxFBBitmapProperty::ChildChanged(wxVariant& thisValue, const int child
...
+					break;
+				}
+				// 'Load From Art Provider'
+                case 5:
```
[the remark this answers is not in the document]
sodevel: Dealing with formatting is such a tedious work here, i have to disable pretty much all auto formatting or Visual Studio reformats half the screen because it wants to correct the surroundings. Even worse when i let it pick up the clang-format file, then it reformats the whole file up to the changed line. Plus the format the clang-format definition produces doesn't look like the format you want to have.

sodevel:
About the localization changes of `_()` to `wxT()`, i don't think i should start with that in this PR, i mean whole wxFB is full of that, i think it would be a better solution to create an own PR just for that issue and correct them all inside that.

src/rad/inspector/wxfbadvprops.cpp
jhasse: I see. What happens when `childVals.GetCount() == 1`?

src/rad/inspector/wxfbadvprops.cpp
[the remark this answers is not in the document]
jhasse: > Plus the format the clang-format definition produces doesn't look like the format you want to have.

The `.clang-format` file uses tabs and has `IndentCaseLabels: true`. What difference do you mean?

jhasse:
Okay, makes sense.

Btw: `wxT()` isn't needed anymore, I think since wxWidgets 2.9. Might still make sense to add it though, as it makes the conversion explicit. I'm not sure about that :/

src/rad/inspector/wxfbadvprops.cpp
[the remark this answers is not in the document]
sodevel: The braces are java-style, the opening brace is on the same line, but you want it on the next line.

sodevel:
This is what they say, but they also put a big limitation on this which is especially a big problem on windows.

Take a look here [Unicode Support in wxWidgets](http://docs.wxwidgets.org/trunk/overview_unicode.html):

> Notice that the narrow strings used with wxWidgets are always assumed to be in the current locale encoding, so writing
> wxMessageBox("Salut à toi!");
> 
> wouldn't work if the encoding used on the user system is incompatible with ISO-8859-1 (or even if the sources were compiled under different locale in the case of gcc). [ ... ] Thus it is important to never use 8-bit (instead of 7-bit) characters directly in the program source but use wide strings

Windows doesn't support UTF-8 as narrow strings so to be on the safe side you must not use anything else than ASCII in a string literal (so no à or ä or whatever). The `wxT()` macro however expands to a wide string, so you are free to use anything that unicode offers and it does work (I stripped the linux part from the quote because its not correct. This charset stuff is complicated, i had much fun with that already).

src/rad/inspector/wxfbadvprops.cpp
sodevel: This raises the urgend need inside me to adjust the other cases as well, after all they are all copy-adjust-pasted and it just doesn't look right if i only change this one but not the other ones...

src/rad/inspector/wxfbadvprops.cpp
sodevel: Good question. Currently this can't happen because each case adds at least one child so there are always at least 2 child elements, but otherwise it looks like the previous value gets kept.

jhasse:
> This is what they say, but they also put a big limitation on this which is especially a big problem on windows.

We're only using ASCII, so there should be no problems.

For UTF-8 literals you can use `wxString::FromUTF8("Salut à toi!");`. `wxT` won't work in this case, as it expects the local encoding, not UTF-8.

> Any update on this?

I'll have a final look and then merge this. Sorry for keeping you waiting.

jhasse:
Thanks!

Issue #429, opened 2018-07-14
wxFALLTHROUGH’ was not declared in this scope


IamSierraCharlie:
Hi,
Followed the instructions in Readme.md

```
git clone --recursive --depth=1 https://github.com/wxFormBuilder/wxFormBuilder
cd wxFormBuilder
./create_build_files4.sh
cd build/3.0/gmake
make config=release
```

Have the following error on Ubuntu 16.04

xrcpreview.cpp
annoyingdialog.cpp
encodingutils.cpp
filetocarray.cpp
stringutils.cpp
typeconv.cpp
wxfbipc.cpp
md5.cc
In file included from /usr/include/wx-3.0/wx/propgrid/propgrid.h:26:0,
                 from /usr/include/wx-3.0/wx/propgrid/manager.h:18,
                 from ../../../src/utils/../rad/inspector/objinspect.h:35,
                 from ../../../src/utils/typeconv.cpp:30:
/usr/include/wx-3.0/wx/propgrid/property.h: In member function ‘void wxPGChoices::Set(const wxArrayString&, const wxArrayInt&)’:
/usr/include/wx-3.0/wx/propgrid/property.h:1049:22: warning: the compiler can assume that the address of ‘values’ will always evaluate to ‘true’ [-Waddress]
         if ( &values )
                      ^
/usr/include/wx-3.0/wx/propgrid/property.h:1049:22: warning: the compiler can assume that the address of ‘values’ will always evaluate to ‘true’ [-Waddress]
../../../src/utils/typeconv.cpp: In function ‘void TypeConv::ParseBitmapWithResource(const wxString&, wxString*, wxString*, wxSize*)’:
../../../src/utils/typeconv.cpp:529:5: error: ‘wxFALLTHROUGH’ was not declared in this scope
     wxFALLTHROUGH;
     ^~~~~~~~~~~~~
wxFormBuilder.make:319: recipe for target 'obj/Release/wxFormBuilder/typeconv.o' failed
make[1]: *** [obj/Release/wxFormBuilder/typeconv.o] Error 1
make[1]: *** Waiting for unfinished jobs....
Makefile:44: recipe for target 'wxFormBuilder' failed
make: *** [wxFormBuilder] Error 2

I've installed this before without issue, but am doing this from scratch again.  I can't find anything relating to wxFALLTHROUGH’ was not declared in this scope type issue although I could be searching for the wrong thing... any ideas?

jhasse:
[...]

PR #440, opened 2018-07-24
Add define for wxFALLTHROUGH for wxWidgets earlier than v3.0.3.

DavidGriffith:
Versions of wxWidgets earlier than v3.0.3 lack a define for wxFALLTHROUGH, which is why #429 happens.  Looking at the source code of wxWidgets, I see this:

```
/* wxFALLTHROUGH is used to notate explicit fallthroughs in switch statements */
#ifdef __cplusplus
#if __cplusplus >= 201103L && defined(__has_warning)
    #if WX_HAS_CLANG_FEATURE(cxx_attributes)
        #define wxFALLTHROUGH [[clang::fallthrough]]
    #endif
#endif
#endif /* __cplusplus */

#ifndef wxFALLTHROUGH
    #define wxFALLTHROUGH ((void)0)
#endif
```

I just opted to add the last ifdef-define-endif, which returns the code to how it was before a3f139f5e20949e9a5a23cc954234a1bacdd10f2.

### DIFF
--- a/src/utils/typeconv.cpp
+++ b/src/utils/typeconv.cpp
@@ -32,6 +32,10 @@
 #include <wx/artprov.h>
 #include <wx/filesys.h>
 
+#ifndef wxFALLTHROUGH
+    #define wxFALLTHROUGH ((void)0)
+#endif
+
 ////////////////////////////////////
 
 // Assuming that the locale is constant throughout one execution,


sodevel:
It is not really wise to define it there, if this symbol gets started to be used elsewhere you have to repeat this definition there as well. Should be defined in a more central header that preferrably already gets included everywhere.

DavidGriffith:
I don't think there is a header that all .cpp include.  Shall I add something like src/common.h for this?

jhasse:
Why not simply update to 3.0.3?

DavidGriffith:
Because Debian 9 still hasn't updated to that version and some people cannot, for various reasons, do an upgrade.

jhasse:
Debian stable will never update a package to include non-security bugfixes or features, but instead forces upstream projects like us to include workarounds. It's pretty selfish and causes lots of pain when people still use Debian stable from 5 years ago and demand that your software compiles with GCC 4.

> for various reasons

I will give you the number 1 reason: They are lazy and want us to workaround their outdated OS.

DavidGriffith:
Okay, so I'll then drop this pull and advise those in this situation to add the define manually.

NewPagodi:
When you want to use wxFALLTHROUGH, can you wrap it in a version check macro?  Like so:

    case <insert case here>:
        #if wxCHECK_VERSION(3,1,0)
            wxFALLTHROUGH;
        #endif

That way if the wxWidgets library version doesn't support wxFALLTHROUGH, the version check macro will keep it from getting through to the compiler and causing an error.

PR #441, opened 2018-07-26
Accept additional wxPanel constructor arguments

### LINKED ISSUE #261: Python __init__ arguments
What starting a new form, the following code is generated for the constructor in Python:

```
class MyPanel1 ( wx.Panel ):
        def __init__( self, parent ):
                wx.Panel.__init__ ( self, parent, id = wx.ID_ANY, pos = wx.DefaultPosition, size = wx.Size( 500,300 ), style = wx.TAB_TRAVERSAL )
```

If I want to be able to set the size, style, or any other panel parameter (some of which can only be set in the constructor), then it would be much more helpful if this was the code generated instead:

```
class MyPanel1 ( wx.Panel ):
        def __init__( self, parent, id = wx.ID_ANY, pos = wx.DefaultPosition, size = wx.Size( 500,300 ), style = wx.TAB_TRAVERSAL, *args, **kwargs ):
                wx.Panel.__init__ ( self, parent, id = id, pos = pos, size = size, style = style, *args, **kwargs )
```

Any other parameter that is currently set by altering the superclass constructor, such as the window_name property, should also be moved to the subclass constructor, or moved to a class method call if available.

Please consider adding this change.

acolomb:
This PR brings the Python and PHP code generation templates up to par with the C++ template.

Add arguments `id`, `pos`, `size`, `style` and `name` to the generated constructors for wxPanel forms. Provide the attributes from the project as default values to these constructor arguments.  Pass the arguments down to the respective parent constructor.

* For Python, this fixes #261.
* The PHP and C++ changes have not been tested at runtime, only checked that the generated source files look correct.
* For C++, only the `#ifnotnull $window_name` check is removed, since the default value is correctly declared as `wxEmptyString` in case the attribute was empty.  This is consistent with the other two language implementations.
* I don't know Lua at all and as far as I could tell, the generated code is not really object oriented like the other languages.  If needed, someone with more Lua-fu might have to look into this.

This allows more flexible use of the generated Panel form classes, especially when reusing the panel within other project elements via their subclass attribute.  That previously failed because the constructors did not accept the extra arguments provided from the project.

### DIFF
--- a/output/plugins/forms/xml/forms.cppcode
+++ b/output/plugins/forms/xml/forms.cppcode
@@ -231,18 +231,18 @@ Written by
 			@{ public wxPanel @}
 		</template>
 		<template name="cons_decl">
-			$name( wxWindow* parent, wxWindowID id = $id, const wxPoint&amp; pos = $pos, const wxSize&amp; size = $size, long style = $window_style #ifnotnull $window_name @{, const wxString&amp; name = $window_name @} );
+			$name( wxWindow* parent, wxWindowID id = $id, const wxPoint&amp; pos = $pos, const wxSize&amp; size = $size, long style = $window_style, const wxString&amp; name = $window_name );
 			#ifequal $aui_managed "1"
 			@{ wxAuiManager m_mgr; #nl @}
 		</template>
 		<template name="cons_def">
-			$name::$name( wxWindow* parent, wxWindowID id, const wxPoint&amp; pos, const wxSize&amp; size, long style#ifnotnull $window_name @{, const wxString&amp; name@} ) :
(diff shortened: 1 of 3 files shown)


acolomb:
(For C++) If the `window_name` attribute was empty in the project file, the generated class constructor did not even have a `name` parameter, in contrast to the default `wxWindow` constructor.  So it could only be set after instantiation by calling `SetName()`.  I'd find that quite surprising because a regular `wxWindow` does accept the constructor parameter, and for wxFB-generated classes, the class interface should not depend upon a specific attribute being set or not IMHO.

I'm not quite sure what the original intention was for the `#ifnotnull` check, but since `wxEmptyString` is used correctly when encountering an empty window_name, it's not necessary at all.

jhasse:
Ah, I think I understand. Thanks for the PR and especially the commit messages explaining your changes :)

Issue #436, opened 2018-07-19
C++ code generation omits some forward declarations


SusanneSchneider:
Hi,
Steps to reproduce:
- Load working fbp-project with subclasses using forward declarations
- Do not change anything, just (re-)generate code
- Some of the previously existing forward declarations are missing, but not all
Some additional info:
- this behaviour was observed in versions 3.6.2 to 3.8. , earlier ones not checked
- related base classes were wxGrid, wxDataViewTreeCtrl, wxTreeListCtrl
- tested with wxWidgets 3.1.0 and 3.1.1
If you have additional questions or need more data, feel free to ask.
Best regards,
Susanne

jhasse:
> previously existing

Which wxFormBuilder version added these forward declarations?

We've added some checks to remove unneeded forward declarations (i.e. the header of the class is also included). Are you forward declarations necessary for your code to compile?

SusanneSchneider:
Hi,
the forward declarations were added by version 3.6.0. Indeed the headers for the missing forward declarations are included, but not explicitly in the wxFormBuilder configurations, but only indirectly via other classes, in the generated projectfb.cpp file. But because these includes are placed in the *.cpp file only, there are compiler errors for the member declarations in the header file. The indirect includes could be a part of the problem, I guess, if wxFormBuilder is checking the forward class names vs. the content of already included headers, but this check does not differentiate between header and source file includes, for example.
Because of the indirect includes (not the cleanest program style, of course....) we currently cannot omit the forward declarations completely.
Regards, Susanne

SusanneSchneider:
[ForwardMissing_example.zip](https://github.com/wxFormBuilder/wxFormBuilder/files/2227894/ForwardMissing_example.zip)
This is an example for reproducing the issue. It's a dialog containing several buttons with different subclasses for each of them.  It contains 3 parts, one created by version 3.6.0, second one after converting this project to version 3.8.0 and last one adding a 3rd button with the default subclass settings of 380 to the converted project. The C++ code generated after conversion from 360 to 380 omitts the forward declarations in the header file. For testing I then added a 3rd button with default settings in version 380, for that button the code was generated correctly. If you check the subclass settings for the default class you see that parameter header is set to forward_declare by default, while it is not set for the buttons converted from version 360.
If you add forward_declare manually, everything seems to work.
But then you have both, the include _and_ the forward setting, and I guess this was not intended by the change.

PR #442, opened 2018-07-26
Fix reading and writing of the subclass element and all other PT_PARENT

sodevel:
The method used to extract the value of a composed wxPGProperty always returns an empty result, only GenerateComposedValue() does return a value. All locations testing a boolean of such a composed property now use the same approach, compare against the property name. Because of the rather strange storage format of a composed boolean and especially the difference to the format of a non-composed boolean comments about the storage format have been added. Parsing the default values of PT_PARENT child elements from a XML file now works properly with empty child tags.

This should fix at least #436


jhasse:
> This should fix at least #436

You mean the missing property when converting from 3.6 to 3.8?

sodevel:
It's not missing, the conversion doesn't touch it according to the zip file. It got read wrong and for sure got damaged when you saved the project because the property grid emptied the value.

PR #446, opened 2018-07-27
Fix eventhandler

sodevel:
Fix assertions because of lost event handlers

The smart trick of the Component base class to pop the event handler if it doesn't match the object does not work because certain wxWidgets elements push event handlers themselves on child objects so this condition cannot be used to determine if the Component itself has pushed one. Remove the cleanup code from the base class and implement it in the derived class instead, this one knows for sure if it has pushed an event handler or not.

NOTE: This introduces an API break, plugins are now required to clean up their pushed event handlers, the base class doesn't do it anymore.

### DIFF
--- a/plugins/additional/additional.cpp
+++ b/plugins/additional/additional.cpp
@@ -834,6 +834,15 @@ public:
 		return grid;
 	}
 
+	void Cleanup(wxObject* obj) override
+	{
+		auto* window = wxDynamicCast(obj, wxGrid);
+		if (window)
+		{
+			window->PopEventHandler(true);
+		}
+	}
+
 	ticpp::Element* ExportToXrc(IObject* obj) override {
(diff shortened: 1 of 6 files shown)


plugins/additional/additional.cpp
```diff
@@ -2046,6 +2136,17 @@ class RibbonPageComponent : public ComponentBase
 
 		return rbpage;
 	}
+
+	/*
```
jhasse: Why add this code comment?

jhasse:
Nice! Does this fix #448?

plugins/additional/additional.cpp
sodevel: These components contain commented event handler pushes so i added commented event handler pops to stay in sync.

sodevel:
Yes it does

jhasse:
Awesome, thank you very much.

For the future, please use `//` to comment multiple lines (multi line `/* */` comments can mess-up automatic merges since affected lines aren't touched which is also bad for `git blame`).

PR #447, opened 2018-07-28
Improve highlight

sodevel:
This improves the drawing of the selection highlights slightly and fixes a possible nullpointer crash.

It adds another interesting feature that was originally planned to fix the disappearing sizer highlight around a wxStaticBoxSizer but has some side effects in the current state. In most situations it currently reenables the highlights inside a wxStaticBoxSizer, but because the wrong parent is used to draw on, depending on the layout, the highlights can be in the wrong position. And because of the missing paint event handler, if the wxStaticBoxSizer is e.g. inside a wxScrolled, parts of the highlights will remain if it is scrolled.

These side effects will disappear with my upcoming PR that enables the currently disabled wxStaticBoxSizer fixes, but that one depends on #446 so i haven't pushed it yet.


sodevel:
This one actually breaks #446 without conflict, will reopen at a later time

PR #449, opened 2018-08-01
Rename wx*_CENTRE flags and update project file

sodevel:
This renames the remaining wx*_CENTRE flags to the new wx*_CENTER flags and also converts the project file accordingly. This is also an addendum to #444 which did not convert the project file according to its changes.


jhasse:
If we bump the projectfile version, I think we should wait until 3.9.0.

sodevel:
Well, this is kind of a problem with the PR workflow approach. Each PR should represent a complete feature itself, but a lot of features require an update of the version number to prevent data loss, like #431 and #432 also do but currently don't.

If each PR contains an update you get a fast growing number, but each PR would be complete and master would always be stable if you merge them. Without the update in the PR, if you merge the PR master wouldn't be stable until next release, but how to handle the outstanding version updates?

You could spilt each PR into two, one with the feature, one with the update. But the outstanding update PR's would all conflict with each other, because master would never contain the version updates from previous PR's and each would base of the latest master version.

And without the split all PR's would conflict because of the same reason.

jhasse:
> lot of features require an update of the version number to prevent data loss

Like? Is it safe to release 3.8.1 from the `master` branch now or should I revert #444 prior to that?

sodevel:
Depends how safe you want it.

Commit dcd659169e2c10fcff63e8c9e06c2ac38b1a4f82 will fix itself, the value will get removed either if you just save the project or if you edit the style flags of the object (i don't know exactly when the style value will get reevaluated). Commit 99e0708c07427d5dd3a3a51c9e0a5100cf7192c7 will cause a loss of the alignment if `wxALIGN_CENTRE` is used (again, don't know exactly when this will happen).

Commit 6fc496011f303aec01b39cf9e367ee2fb8d9ef58 only adds some values to a present attribute, this won't cause a problem on upgrade, but these flags will silently be lost if you open the project file with an older version of wxFB. Commit efdc5d3b69d84795302b82e10988a78c09bb428f adds new attributes, this again won't cause a problem on upgrade and older versions of wxFB will at least emit a warning about unknown attributes that will get lost on save, so this is the safest addition.

Sorry for discovering the upgrade issue so late, i wanted to add a fix but you already merged the PR. But i guess i am in good company, in the past the version number was increased only sparingly, most of the time upgrades are working well, but silent data loss because of using a newer project file with an older wxFB was not taken care of.

jhasse:
> Commit 99e0708 will cause a loss of the alignment if wxALIGN_CENTRE is used (again, don't know exactly when this will happen).

Won't this happen every time I open an old project file which uses `wxALIGN_CENTRE`?

sodevel:
Without the increased version number and the update code of this PR yes.

Renaming things is the worst case and always require a new version number so that the update code can convert previous project files. Deleting or adding things isn't that worse and could be done without a new version number. Elements will get deleted or added when the project gets modified, here its just a problem if the project file gets opened with an older wxFB version, it might silently remove new elements again.

jhasse:
So if I understand it correctly: revert #444 for 3.8.1 if we don't want to bump the version?

sodevel:
If you cannot revert only 99e0708c07427d5dd3a3a51c9e0a5100cf7192c7 then you should revert the whole #444, and reopen #407

jhasse:
Reverting only 99e0708 worked. I'll release 3.8.1 this week and then apply the patch again :)

PR #455, opened 2018-08-03
Enable the disabled wxStaticBoxSizer highlight fixes

sodevel:
This re-enables the commented wxStaticBoxSizer highlight fixes from my ancient patch. I don't know why they were disabled, but they do no harm, im using them for years, and because i use a lot of wxStaticBoxSizers it would be nice to get the highlight working inside them again.

It is rather incovenient to add another special treatment for some element, but i currently don't know how to handle this special case in a more common manner easily. The wxStaticBoxSizer is basically two elements, the invisible sizer and the, in the wxFB object hierary, non-existent visible wxStaticBox. The current data model just can't handle this case.

The wxCollapsiblePane also has broken highlights but there the situation is even more complicated because this element is build up of two visible objects, again only one known to wxFB, with one disappearing depending on state.

### DIFF
--- a/src/rad/designer/visualeditor.cpp
+++ b/src/rad/designer/visualeditor.cpp
@@ -103,18 +103,15 @@ void VisualEditor::DeleteAbstractObjects()
 		}
 		else
 		{
-			//TODO: fix wxStaticBox later...
-			/*if( wxStaticBoxSizer* staticBoxSizer = dynamic_cast< wxStaticBoxSizer* >( it->first ) )
+			// Delete push'd visual object event handlers
+			if (auto* staticBoxSizer = wxDynamicCast(it->first, wxStaticBoxSizer))
 			{
-				if( wxWindow* window = staticBoxSizer->GetStaticBox() )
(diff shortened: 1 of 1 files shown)


jhasse:
IIRC `wxDynamicCast` is from a long time ago, where C++ didn't necessarily had RTTI. As this isn't the case today, I would rather use `dynamic_cast`, as it makes it clearer to C++ developers what is happening. Can you change that?

Otherwise it looks good, I'm not using `wxStaticBoxSizer` very often though.

jhasse:
> I think the rationale behind wxWidgets RTTI these days is that it is not so expensive like C++ RTTI.

Hm ... why should that be the case?

> AFAIK its code base uses wxWidgets RTTI whenever possible, thats why i used it in #446 as well.

I see. I can merge this how it is, but let's stick to C++ RTTI for the future.

jhasse:
The dynamic information only gets added to objects with virtual methods.

AFAIK wxWidgets adds its class information to every object derived from wxObject, so you don't have that much control over it.

Issue #283, opened 2017-05-10
error: unknown type name 'wxMediaCtrl'


dicemans:
```
In file included from /usr/local/include/wx-3.0/wx/propgrid/propgrid.h:26:
/usr/local/include/wx-3.0/wx/propgrid/property.h:1049:15: warning: reference cannot be bound to dereferenced null pointer in
      well-defined C++ code; pointer may be assumed to always convert to true [-Wundefined-bool-conversion]
        if ( &values )
        ~~    ^~~~~~
../../../plugins/additional/additional.cpp:1290:3: error: unknown type name 'wxMediaCtrl'
                wxMediaCtrl* mc = new wxMediaCtrl((wxWindow *)parent, wxID_ANY, wxT(""),obj->GetPropertyAsPoint(_("pos")),
                ^
../../../plugins/additional/additional.cpp:1290:25: error: unknown type name 'wxMediaCtrl'
                wxMediaCtrl* mc = new wxMediaCtrl((wxWindow *)parent, wxID_ANY, wxT(""),obj->GetPropertyAsPoint(_("pos")),
                                      ^
../../../plugins/additional/additional.cpp:1303:100: error: use of undeclared identifier 'wxMEDIACTRLPLAYERCONTROLS_STEP'
  ...if(  obj->GetPropertyAsString( _("player_controls") ) == wxT("STEP") )     mc->ShowPlayerControls(wxMEDIACTRLPLAYERCONT...
                                                                                                    ^
../../../plugins/additional/additional.cpp:1304:102: error: use of undeclared identifier 'wxMEDIACTRLPLAYERCONTROLS_VOLUME'
  ...if(  obj->GetPropertyAsString( _("player_controls") ) == wxT("VOLUME") )   mc->ShowPlayerControls(wxMEDIACTRLPL...
                                                                                                            ^
../../../plugins/additional/additional.cpp:1305:103: error: use of undeclared identifier 'wxMEDIACTRLPLAYERCONTROLS_DEFAULT'
  ...if(  obj->GetPropertyAsString( _("player_controls") ) == wxT("DEFAULT") )  mc->ShowPlayerControls(wxMEDIACTRLPL...
                                                                                                            ^
../../../plugins/additional/additional.cpp:1306:100: error: use of undeclared identifier 'wxMEDIACTRLPLAYERCONTROLS_NONE'
[...]

PR #462, opened 2018-08-08
Update README as issue #283 seems to be fixed

muellermartin:


muellermartin:
The build failed due to missing `gmake`. I've added it to the dependencies in the README.
Also the Travis CI script was updated to include the missing dependency and exclude the `--disable-mediactrl` flag.

jhasse:
Thanks :)

PR #493, opened 2018-12-15
Modify python template for BitmapComboBox to comply with changes made in wxPython version 4

jmoraleda:
`BitmapComboBox` has been moved from module `combo` to  module `adv `in wxPython version 4.   (https://wxpython.org/Phoenix/docs/html/wx.adv.BitmapComboBox.html#wx.adv.BitmapComboBox). In fact wxPython 4 does not have a `combo` module`.

Also `BitmapComboBox` default `choices` argument should be an empty list `[]` and not an empty string `""`. 

Without both of these changes python code for user interfaces that include `BitmapComboBox` will produce an error when run using wxPython version 4.

### DIFF
--- a/output/plugins/common/xml/common.pythoncode
+++ b/output/plugins/common/xml/common.pythoncode
@@ -117,7 +117,7 @@ Python code generation written by
 
   <templates class="wxBitmapComboBox">
 	<template name="construction">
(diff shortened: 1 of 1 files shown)


jhasse:
Thanks!

Issue #204, opened 2014-04-16
Relative paths not work


wxfbint:
Hello,
I found another bug in the last version from PPA in Ubuntu 14.04 (same behavior in Ubuntu 13.10)
wxFB create the same paths for the images, no matter what I select in the project in the property "relative_path" check or uncheck it.
Im attaching a sample fbp file with the project property relative_path= unchecked you can see the Python code (line 26) created, with the property checked is the same code.

Reported by: mlacunza

jhasse:
Fixed by #496.

Issue #104, opened 2010-12-23
wxGridBagSizer - colspan issue


wxfbint:
etting the colspan of a wxGridBagSizer to 0 will hang the application and make it eat cpu/memory.

Sorry, I forgot to mention, that this is the (gbsizeritem) child, not the wxGridBagSizer control itself

Reported by: jfouche

jhasse:
Fixed by #496.

Issue #78, opened 2010-12-23
Add images relative to wxTheApp->GetPath()


wxfbint:
I have a problem with the files loaded from relative paths and bitmap buttons/images.

Imagine this situation:

\- You create a new dialog in wxFB. You place a bitmap button with an image “./MyImage.png”. You create the .H.

\- In your application, you use the .H created by wxFB.

\- Then, the application asks the user to select a file using a wxFileDialog call. The current folder changes...

\- Now, I open the wxFB dialog using the “./MyImage.png”... and voilá! A file not found assert is fired... That’s because the current folder changed and, like I specified a “./MyImage.png” as bitmap button’s icon then the error appears correctly...

SOLUTION: Add an option in wxFB to use relative-to-application image paths. For example, I could specify for the bitmap button created in wxFB a path like “{AppPath}/MyImage.png”. Then, the dialog’s constructor on your wxFB .CPP output can call something like

this-&gt;MyButton-&gt;SetIcon(wxTheApp-&gt;GetPath()+wxT("/MyImage.png"));
Thanks.

Reported by: jfouche

jhasse:
Fixed by #496.

Issue #75, opened 2010-12-18
Bitmap relative path


wxfbint:
Migrating this bug from the old bug tracker.

I have my source located in a directory called "Source", my final executable builds to a directory called "Bin", and images are found in "Bin/Data/Images"; this means that any paths in the automatically generated code look like "../Bin/Data/Icons/" due to it exporting the paths relative to the code export location.

This actually works ok providing that the executable is ALWAYS in a directory called "Bin", my suggestion would have been to have two paths, one for where you wanted the exported code to be sent (in my case "Source"), and another for your executables working directory (in my case "Bin"),

Then the exported code could use wxTheApp-&gt;GetPath() followed by the image location relative to the working directory set.

Example:
working path = Bin/
image path = Bin/Data/Images/add.png
Exported code: wxTheApp-&gt;GetPath() + wxT("/Data/Images/add.png")

Reported by: *anonymous

jhasse:
Fixed by #496.

PR #496, opened 2019-01-29
Python: Add wrapper function to bitmap paths for easier runtime-path …

ANoDE85:
…configuration. Addresses #204 #104 #75 #78 

The idea behind this is, to let the user specify a function name in the Python settings. When generating python code, bitmap paths will be wrapped into this function call:

```
class MyFrame1 ( wx.Frame ):

	def __init__( self, parent ):
		# [...]
		self.m_bitmap1 = wx.StaticBitmap( self, wx.ID_ANY, wx.Bitmap( self.GetRuntimeImagePath( u"bitmap.png" ), wx.BITMAP_TYPE_ANY ), wx.DefaultPosition, wx.DefaultSize, 0 )
		# [...]

	# Virtual image path resolution method. Override this in your derived class.
	def GetRuntimeImagePath( self, bitmap_path ):
		return bitmap_path
```

This would allow for manipulating the runtime image paths via subclassing.


jhasse:
Thanks! It seems that you have changed the line endings, making the diff unuseable on GitHub. Can you move those changes into a separate commit (or drop them)?

ANoDE85:
> Thanks! It seems that you have changed the line endings, making the diff unuseable on GitHub. Can you move those changes into a separate commit (or drop them)?

Hi,
sorry, that was a mistake. My GIT was configured in an incorrect way. I added another commit fixing the eol-changes. I hope that works for you. (I'm REALLY inexperienced when it comes to Git / GitHub, so sorry if I appear stupid).

jhasse:
Thanks!

PR #500, opened 2019-02-08
Interpret a missing child value of type boolean as true

sodevel:
The subclass property is currently the only one using this child type, because the only instance using this property, the c++ code generator, interprets a missing value as true and currently no project file update
adds this value if it is missing, the object inspector also needs to interpret a missing value as true to stay in sync.

### DIFF
--- a/src/rad/inspector/objinspect.cpp
+++ b/src/rad/inspector/objinspect.cpp
@@ -492,7 +492,12 @@ void ObjectInspector::AddItems( const wxString& name, PObjectBase obj,
 							// Because the format of a composed wxPGProperty value is stored this needs to be converted
 							// true == "<property name>"
 							// false == "Not <property name>"
-							child = new wxBoolProperty(it->m_name, wxPG_LABEL, value == it->m_name);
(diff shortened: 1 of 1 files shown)


jhasse:
What is left "TODO"?

sodevel:
On first thought i had the impression it is just enough to introduce a new project file version to be able to update the old values to contain the third value, but on second thought i wonder why the default value of the element description is not used.

As far as i remember all elements defined in the XML files contain a default value that is used if the object in the project file doesn't have a value for the element, but why isn't it used here? This PT_PARENT element is kind of special, the values of its children is used to build a compound value that gets parsed in certain locations, because the project files with only two elements inside this compound value do have that compound value, maybe this one is parsed and no one looks at the default value of the compound value and the missing child value stays empty?

Sadly it's been a while i was working on the codebase and forgot how this works and it's not so easy to understand, so as a quick fix i reintroduced that empty-check that funny enough one of my previous PR removed :smile:

jhasse:
I see, thanks for the explanation :)

PR #511, opened 2019-03-28
Lua code gen overhaul

paul-reilly:
This is a big change to how Lua code gen works.

- There is no more global table, it produces a module.
- It allows instantiating when the user wants to, not on first parse
- It has a callback system for events where empty functions are called, which can be overridden later without touching the autogenerated code.

Usage:

    local ui = require("gui")

    ui.MainFrame:create()
    ui.MainFrame.MainFrame:Show()

    -- or

    local ui = require("gui")

    local main_frame = ui.MainFrame:create()
    main_frame:Show

All children are still accessible through the form entries in the table.

Let me know what you think, about anything, or if any bugs are found. I also enabled auiToolBar now that it's implemented in wxLua.

### DIFF
--- a/src/codegen/luacg.cpp
+++ b/src/codegen/luacg.cpp
@@ -1264,7 +1264,7 @@ void LuaCodeGenerator::GenConstructor(PObjectBase class_obj, const EventVector& 
 
 	wxString self("self.");
 	GenEvents( class_obj, events, strClassName );
-	m_source->WriteLn("return " + controlName);
+	m_source->WriteLn("return self." + controlName);
 	m_source->Unindent();
 	m_source->WriteLn("end");
 }
--- a/src/codegen/luacg.cpp
+++ b/src/codegen/luacg.cpp
@@ -1362,8 +1362,8 @@ void LuaCodeGenerator::GenConstruction(PObjectBase obj, bool is_widget, wxString
 				else
 					parentPostfix = wxEmptyString;
 
-				_template = wxT( " self.#parent$name" ) + parentPostfix + wxT( ":SetSizer( .self$name ) #nl" )
-					    wxT( "self.#parent$name" ) + parentPostfix + wxT( ":Layout()" )
+				_template = wxT( " self.#parent$name" ) + parentPostfix + wxT( ":SetSizer( self.$name ) #nl" )
+					    wxT( " self.#parent$name" ) + parentPostfix + wxT( ":Layout()" )
 					    wxT( "#ifnull #parent $size" )
 					    wxT( "@{ #nl self.$name:Fit( self.#parent $name" ) + parentPostfix + wxT( " ) @}" );
 
--- a/src/codegen/luacg.cpp
+++ b/src/codegen/luacg.cpp
@@ -40,6 +40,11 @@
 #include <wx/filename.h>
 #include <wx/tokenzr.h>
 
+// hack to replace some "UI"s with the new "self"s since the source is unclear (to me)
+static void FindReplaceUIWithSelf(wxString& code) {
+	if (code.Find("UI") == 0) code.Replace("UI", "self");
+}
+
 LuaTemplateParser::LuaTemplateParser( PObjectBase obj, wxString _template, bool useI18N, bool useRelativePath, wxString basePath, std::vector<wxString> strUserIDsVec )
 :
 TemplateParser(obj,_template),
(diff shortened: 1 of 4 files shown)


mojomojomojo:
Yes! I like wxFormbuilder a lot, and this was the reason I had avoided it for lua projects. Then
Thank you.

paul-reilly:
I've found a bug, so will try to fix it.

src/codegen/luacg.cpp
```diff
@@ -1902,4 +1922,4 @@ void LuaTemplateParser::SetupModulePrefixes()
 
 	ADD_PREDEFINED_PREFIX( wxAC_DEFAULT_STYLE, wx. );
 	ADD_PREDEFINED_PREFIX( wxAC_NO_AUTORESIZE, wx. );
-}
+}
```
jhasse: Please keep a newline at the end of the file here.

paul-reilly: No worries, will do.

src/codegen/luacg.cpp
```diff
@@ -612,11 +617,8 @@ bool LuaCodeGenerator::GenerateCode( PObjectBase project )
 	PProperty propNamespace = project->GetProperty( wxT( "ui_table" ) );
 	if ( propNamespace )
 	{
-		m_strUITable = propNamespace->GetValueAsString();
-		if(m_strUITable.length() <= 0){
-			m_strUITable = wxT("UI");//default value ... m_strUITable shouldn't be empty
-		}
-		code = m_strUITable +  wxT(" = {} \n");
+		m_strUITable = "ui";
```
sodevel: Ignoring the project property and replacing with a hardcoded value, why?

src/codegen/luacg.cpp
```diff
@@ -866,7 +875,7 @@ wxString LuaCodeGenerator::GetCode(PObjectBase obj, wxString name, bool silent /
 
 	if (!m_strUITable.empty())
 	{
-		_template.Replace(wxT("#utbl"), m_strUITable + wxT("."));
+		_template.Replace(wxT("#utbl"), wxT("self."));
```
sodevel: And if the table name is empty, the empty string will be used, doesn't look right

src/codegen/luacg.cpp
```diff
@@ -949,7 +958,7 @@ wxString LuaCodeGenerator::GetConstruction(PObjectBase obj, bool silent, wxStrin
 		strTableName = table->GetValueAsString();
 		if (strTableName.empty())
 		{
-			strTableName = wxT("UI");
+			strTableName = "self";
```
sodevel: If the table name is not empty, its value will be used, not this new default value

src/codegen/luacg.cpp
```diff
@@ -1021,6 +1030,7 @@ void LuaCodeGenerator::GenClassDeclaration(PObjectBase class_obj, bool /*use_enu
 	}
 
 	GetGenEventHandlers( class_obj );
+	m_strUITable = "ui";
```
sodevel: This member should be set only once, it contains the project property of the table name. Here it gets replaced with a hardcoded value and never gets restored

src/codegen/luacg.cpp
```diff
@@ -1208,11 +1218,17 @@ void LuaCodeGenerator::GenConstructor(PObjectBase class_obj, const EventVector&
...
+	m_source->WriteLn(strName + " = Object()");
+	// functions are added to a local 'ui' table and are named the same as the control
+	// name, eg ui.MainFrame
+	m_source->WriteLn("function " + strName + ":create()");
 	m_source->Indent();
+	wxString controlName = strName;
+	controlName.Replace(m_strUITable + ".", "");
+	m_source->WriteLn("self." + controlName + " = " + GetCode(class_obj, wxT("cons_call")));
+
+	m_strUITable = "self";
```
sodevel: Again replacing this member with another hardcoded value and never restored

paul-reilly:
Thanks for the feedback, I agree that it was submitted in non-optimum shape, I'm also now aware of the source of the `ui_table`, "UI", which only needs to be customisable using the old way since the user can now name the imported module on import. I have now removed the `FindReplaceUIWithSelf` hack and have made various other tidies. There's still a bug or two though.

I was/am operating under the assumption that almost nobody uses the old version because although it lays some fantastic groundwork for Lua code gen, the code model isn't correct. For example, if you define dialog windows, they all show on application start, it doesn't follow the code gen model of never having to touch the code and there's zero encapsulation between names in different forms.

I'll work on it on my fork over the next while and see where it goes.

PR #512, opened 2019-03-28
correct python generated code

seandepagnier:


sodevel:
According to the documentation this is not a fix but kind of a revert. The current template generates code for Phoenix but this PR reverts it to Classic. Since Phoenix is the current wxPhython implementation i think this should not be merged.

jhasse:
Yep, Phoenix is the way to go ;)

PR #521, opened 2019-06-09
wxCollapsePane bug  and StyledTextCtrl for python

JacoboSpain:
-> Add StyledTextCtrl for python.
-> BUG generate code for PT_WXPARENT_CP, missing self

### DIFF
--- a/output/plugins/additional/xml/additional.pythoncode
+++ b/output/plugins/additional/xml/additional.pythoncode
@@ -322,9 +322,7 @@ Python code generation written by
 			#nl self.$name.MarkerDefine( wx.stc.STC_MARKNUM_FOLDERMIDTAIL, wx.stc.STC_MARK_EMPTY )
 			#nl self.$name.MarkerDefine( wx.stc.STC_MARKNUM_FOLDERTAIL, wx.stc.STC_MARK_EMPTY )
 			#nl self.$name.SetSelBackground( True, wx.SystemSettings.GetColour(wx.SYS_COLOUR_HIGHLIGHT ) )
-			#nl self.$name.SetSelForeground( True, wx.SystemSettings.GetColour(wx.SYS_COLOUR_HIGHLIGHTTEXT ) )
-			<!-- @# WARNING: wxPython code generation isn't supported for this widget yet. #nl
-		    self.$name = wx.Window( #wxparent $name )-->			
+			#nl self.$name.SetSelForeground( True, wx.SystemSettings.GetColour(wx.SYS_COLOUR_HIGHLIGHTTEXT ) )		
 		</template>
 		<template name="include">import wx.stc</template>
 	</templates>
--- a/output/plugins/additional/xml/additional.pythoncode
+++ b/output/plugins/additional/xml/additional.pythoncode
@@ -269,9 +269,64 @@ Python code generation written by
 
 	<templates class="wxStyledTextCtrl">
 		<template name="construction">
-			@# WARNING: wxPython code generation isn't supported for this widget yet. #nl
-			self.$name = wx.Window( #wxparent $name )
+			self.$name = wx.stc.StyledTextCtrl(#wxparent $name, $id,  $pos, $size, $window_style)
+			#nl self.$name.SetUseTabs  ( $use_tabs )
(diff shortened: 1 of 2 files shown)


output/plugins/additional/xml/additional.pythoncode
```diff
@@ -269,9 +269,64 @@ Python code generation written by
...
+			#nl self.$name.MarkerSetBackground( wx.stc.STC_MARKNUM_FOLDEREND, wx.BLACK )
+			#nl self.$name.MarkerSetForeground( wx.stc.STC_MARKNUM_FOLDEREND, wx.WHITE )
+			#nl self.$name.MarkerDefine( wx.stc.STC_MARKNUM_FOLDEROPENMID, wx.stc.STC_MARK_BOXMINUS )
+			#nl self.$name.MarkerSetBackground( wx.stc.STC_MARKNUM_FOLDEROPENMID, wx.BLACK)
+			#nl self.$name.MarkerSetForeground( wx.stc.STC_MARKNUM_FOLDEROPENMID, wx.WHITE)
+			#nl self.$name.MarkerDefine( wx.stc.STC_MARKNUM_FOLDERMIDTAIL, wx.stc.STC_MARK_EMPTY )
+			#nl self.$name.MarkerDefine( wx.stc.STC_MARKNUM_FOLDERTAIL, wx.stc.STC_MARK_EMPTY )
+			#nl self.$name.SetSelBackground( True, wx.SystemSettings.GetColour(wx.SYS_COLOUR_HIGHLIGHT ) )
+			#nl self.$name.SetSelForeground( True, wx.SystemSettings.GetColour(wx.SYS_COLOUR_HIGHLIGHTTEXT ) )
+			<!-- @# WARNING: wxPython code generation isn't supported for this widget yet. #nl
```
sodevel: You implemented code generation for this control, no need to keep the old placeholder code, just remove it

src/codegen/pythoncg.cpp
```diff
@@ -24,6 +24,9 @@
 // Python code generation writen by
 //   Michal Bližňak - michal.bliznak@gmail.com
 //
+//   Jacobo Costas Costas
```
sodevel: Don't add credits for small fixes or additions. So far these credits have only been used on large additions like support for a new generator language.

src/codegen/pythoncg.cpp
```diff
@@ -102,7 +105,7 @@ wxString PythonTemplateParser::ValueToCode( PropertyType type, wxString value )
 		}
 	case PT_WXPARENT_CP:
 	{
-		result = value + wxT(".GetPane()");
+		result = wxT("self.")  + value + wxT(".GetPane()");
```
sodevel: What are you trying to fix with this? I don't think this is correct, i did a quick test with the current version and the `self.` gets added in the proper places without your change, with your change this should result in `self.self.` in the places.

JacoboSpain: The problem is when elements are inserted into the collapsePane, and it is necessary to call GetPane ()

For example: With original trunk a have a frame:
Frame

The program generate a code:
fail_collapsePane

With my coding:

fail_collapsePane

Insert .self when GetPane is called

sodevel: I still can't reproduce your issue, i have rebuild your layout and this is the generated code:

collapse

This is made by wxFB 48f2f0f, which revision are you using?

JacoboSpain: last revision 0e591aaa7a98ca756359bd5dd8cd2cfa9a7d1140
Compile in Windows 10 64 bits with MSYS32 with gcc 9.1.0

sodevel: Ah i see, thats the reason, my revision is too old, it doesn't contain the changes to the code generation that have been made by myself 😄. I can reproduce the issue now and unfortunately not only Python is affected but also PHP and Lua, but for some reason in Lua the error doesn't manifest. In contrast to the wxStaticBoxSizer case the path to the parent wx element needs to get prepended.

This fix is kind of unrelated to the added control, can you please remove it from this PR and file it as a separate one, plus maybe fixing the other two languages as well?

JacoboSpain: ok

PR #527, opened 2019-06-12
Insert StyledTextEditor for python

JacoboSpain:
Insert StyledTextEditor for python

### DIFF
--- a/output/plugins/additional/xml/additional.pythoncode
+++ b/output/plugins/additional/xml/additional.pythoncode
@@ -269,9 +269,62 @@ Python code generation written by
 
 	<templates class="wxStyledTextCtrl">
 		<template name="construction">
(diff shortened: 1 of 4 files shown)


sodevel:
Thank you for cleaning up the commits and adding the additional fixes, but as you maybe discovered already yourself, both commits again are part of the same PR. I really don't want to merge it like that to keep the history clean.

You don't seem to be familiar with git so let me give you some assistance on that topic. You made both commits to the master branch, in general you don't do commits directly to master, instead you work on branches and merge these into master. To get this PR split, at first you have to rewind master in your fork to the same revision like in this repository. From that location you create two branches, one for each PR. You switch into the first branch and commit the new control, you switch to the second branch and commit the fixes. Then you can propose each branch as an individual PR.

JacoboSpain:
OK. Sorry It's my first contribution on github

PR #529, opened 2019-06-13
wxFB bug with collapsePane

JacoboSpain:
solved problem when calling getPane in wxCollapsePane in Python, PHP and Lua


sodevel:
I squashed these commits to remove the changes and their revert from your other PR. For the next time you need to "delete" a commit, it is better to do a rebase to remove it and force push the branch.

Issue #520, opened 2019-06-05
OnCharHook bring the form name instead of "this"


hjnsxx:
The event OnCharHook generate the next line which cause a compiler error:

GUIFrame->Connect(wxEVT_CHAR_HOOK, wxKeyEventHandler(GUIFrame::GUIFrameOnCharHook));

I should change "GUIFrame" with "this" to work

this->Connect(wxEVT_CHAR_HOOK, wxKeyEventHandler(GUIFrame::GUIFrameOnCharHook));

same on "Disconnect"

thanks.

PR #534, opened 2019-08-10
Fix for OnCharHook event

Kvaz1r:
Fix for #520


sodevel:
Thanks for fixing that issue.

sodevel:
Fixes #520

Issue #478, opened 2018-10-19
Compile errors in generated code when using wxStaticBoxSizer::OnUpdateUI


ga2k:
Choosing to implement the **OnUpdateUI** event for a **wxStaticBoxSizer** generates incorrect code that won't compile.

**(In the generated wxPanel ctor)**

`this->Connect( userDetailsFrameSizer->GetId(), wxEVT_UPDATE_UI, wxUpdateUIEventHandler( PanelBase::uiUserDetails ) );`

**(In the build output)**

_1>d:\dev\pbstracker\framebase.cpp(1017): error C2039: 'GetId': is not a member of 'wxStaticBoxSizer'
1>d:\libs\wxwidgets\wxwidgets-3.1.1\include\wx\sizer.h(1033): note: see declaration of 'wxStaticBoxSizer'_

**(In the generated wxPanel dtor)**

`	
this->Disconnect( userDetailsFrameSizer->GetId(), wxEVT_UPDATE_UI, wxUpdateUIEventHandler( PanelBase::uiUserDetails ) );`

**(In the build output)**

_1>d:\dev\pbstracker\framebase.cpp(1053): error C2039: 'GetId': is not a member of 'wxStaticBoxSizer'
1>d:\libs\wxwidgets\wxwidgets-3.1.1\include\wx\sizer.h(1033): note: see declaration of 'wxStaticBoxSizer'_

**Editing the generated code to use the actual id's of the controls works as expected.**

However, editing the generated code is unwise as the entire .cpp and .h files are regenerated constantly.

PR #535, opened 2019-08-13
Fix compile error for wxStaticBoxSizer::OnUpdateUI

Kvaz1r:
Closes:  #478 

Sizer itself hasn't an id and can't been bound to event, so need to use inner StaticBox instead.

### DIFF
--- a/output/plugins/layout/xml/layout.cppcode
+++ b/output/plugins/layout/xml/layout.cppcode
@@ -64,7 +64,7 @@ Written by
 	</template>
     <template name="construction">$name = new wxStaticBoxSizer( new wxStaticBox( #wxparent $name, $id, $label ), $orient ); #nl</template>
 	<template name="evt_entry_OnUpdateUI">EVT_UPDATE_UI( $id, #handler )</template>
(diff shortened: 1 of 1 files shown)


sodevel:
It seems like C++ is not the only language affected by this, PHP and Python templates seem to have the same error. I'm really not familar with these languages nor their wxWidgets port, maybe you know more about them and can fix it there as well?

Also if you reference issues can you please use the proper syntax [https://help.github.com/en/articles/closing-issues-using-keywords](url) so that the issue gets closed automatically when i merge this.

Kvaz1r:
> I'm really not familar with these languages nor their wxWidgets port, maybe you know more about them and can fix it there as well?

Unfortunately I don't know anything about this languages too. I can change by analogue, but I'm not really sure will it better or not. 

> And maybe use a more meaningful name for the branch than patch-1, its name is used in the merge commit message and your last branch also had this name, this looks a bit confusing in the history.

It's automatic name, I will change it next time. 

> Also if you reference issues can you please use the proper syntax
Right, fixed.

Issue #287, opened 2017-06-06
Crash when creating 2 sizers within a frame


TomasRiker:
The [latest successful build](https://ci.appveyor.com/api/projects/jhasse/wxformbuilder-461d5/artifacts/wxFormBuilder_win32.zip?branch=master) allows to create an object hierarchy like this, which is probably invalid and apparently only works using a `wxGridBagSizer` as the second sizer (the program won't allow you to create e.g. two `wxBoxSizer`s directly within the same frame):

![image](https://cloud.githubusercontent.com/assets/295205/26814050/b5241150-4a82-11e7-857e-bd49df8544df.png)

Then, when changing the selection in the Object Tree, the program crashes (Windows dialog "_wxFormBuilder.exe stopped working_").

PR #537, opened 2019-08-20
Add check when create wxGridBagSizer

Kvaz1r:
Fixes: #287 

wxGridBagSizer has own type so there is need to check if there any other sizers and vice versa when adding wxGridBagSizer.

### DIFF
--- a/src/model/database.cpp
+++ b/src/model/database.cpp
@@ -228,6 +228,19 @@ int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent,PObjectType typ
 	return count;
 }
 
+int ObjectDatabase::CountChildrenWithTypes(PObjectBase parent,const std::set<wxString> &types)
+{
+	unsigned int count = 0;
+	unsigned int numChildren = parent->GetChildCount();
+	for (unsigned int i=0; i < numChildren ; i++)
+	{
+		if (types.find(parent->GetChild(i)->GetObjectInfo()->GetObjectType()->GetName()) != types.end())
+			count++;
+	}
+
+	return count;
+}
+
 /**
 * Crea una instancia de classname por debajo de parent.
 * La función realiza la comprobación de tipos para crear el objeto:
@@ -257,7 +270,7 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
 	}
 
 	PObjectType objType = objInfo->GetObjectType();
-
+    
 	if (parent)
 	{
 		// Comprobamos si el tipo es válido
@@ -301,9 +314,23 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
 			bool create = true;
 
 			// we check the number of instances
-			if (max > 0 && CountChildrenWithSameType(parent, objType) >= max)
-				create = false;
+			int count;
+			if (objType->GetName() == wxT("sizer") || classname == "wxGridBagSizer")
+			{
+				std::set<wxString> types;
+				types.insert(wxT("sizer"));
+				types.insert(wxT("gbsizer"));
+				count = CountChildrenWithTypes(parent, types);
+			}
+			else 
+			{
+				count = CountChildrenWithSameType(parent, objType);
+			}
+			
 
+			if (max > 0 && count >= max)
+				create = false;
+				
 			if (create)
 				object = NewObject(objInfo);
 		}
--- a/src/model/database.cpp
+++ b/src/model/database.cpp
@@ -228,14 +228,14 @@ int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent,PObjectType typ
 	return count;
 }
 
-int ObjectDatabase::CountChildrenWithTypes(PObjectBase parent,const std::set<wxString> &types)
+int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent, const std::set<PObjectType>& types)
 {
 	unsigned int count = 0;
 	unsigned int numChildren = parent->GetChildCount();
-	for (unsigned int i=0; i < numChildren ; i++)
+	for (unsigned int i = 0; i < numChildren; ++i)
 	{
-		if (types.find(parent->GetChild(i)->GetObjectInfo()->GetObjectType()->GetName()) != types.end())
-			count++;
+		if (types.find(parent->GetChild(i)->GetObjectInfo()->GetObjectType()) != types.end())
+			++count;
 	}
 
 	return count;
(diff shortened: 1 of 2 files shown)


sodevel:
This code is hard to understand and hides the problem it is trying to fix: instead of counting children of one type you have to count children of two, equivalent, types. This code derives the type from two fixed classes you assume to have a specific type, this is not only confusing but also sort of dangerous and misleading.

I think a much better approch is to specifically test for the types sizer and gbsizer if the target type is one of these. This is a more clear and direct approach.

Kvaz1r:
First of all I was a bit surprised that gbsizer has own type and all other sizers are not, so I just take more common class for getting type. 

> I think a much better approch is to specifically test for the types sizer and gbsizer if the target type is one of these. This is a more clear and direct approach.

Agree, but I don't understand how to do these way(i.e create type just by name). Do you propose change CountChildrenWithSameType function and add check there?

sodevel:
The extra type is probably required because of the extra sizeritem type.

It seems the only way to get an object type item is to query it by name like e.g. some code above already does. I think instead of hiding the extra comparison inside `CountChildrenWithSameType` it would be better to either provide an overload that accepts two types or modify the method to accept something like a set of types to compare against.

Kvaz1r:
ok, I will try update PR tomorrow with set of types as argument because it's only one place where used CountChildrenWithSameType function. 
Should I rename it to CountChildrenWithTypes and use std::set as container?

sodevel:
It seems like this is the only case where two types need to be tested together, please add this method as an overload, this saves the cost of creating the set for the very common case where only one type needs to be tested, use the new method only for the special case of these two sizer types.

src/model/database.h
```diff
@@ -197,6 +197,7 @@ class ObjectDatabase
   PObjectType GetObjectType(wxString name);
 
   int CountChildrenWithSameType(PObjectBase parent,PObjectType type);
+  int CountChildrenWithTypes(PObjectBase parent,const std::set<wxString> &types);
```
sodevel: I think this really should be an overload instead of another method, also formatting

src/model/database.cpp
```diff
@@ -301,9 +314,23 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
 			bool create = true;
 
 			// we check the number of instances
-			if (max > 0 && CountChildrenWithSameType(parent, objType) >= max)
-				create = false;
+			int count;
+			if (objType->GetName() == wxT("sizer") || classname == "wxGridBagSizer")
```
sodevel: wxFB uses only types in its datamodel for these checks so don't test against a class, test against its type.

The object type system uses one object for each type and does identity comparison between them, so don't compare them by their name, just use the name to get the object and compare these. This way you don't have to create a set of `wxString` but a set of `PObjectType`, comparisons among these are cheaper than string comparisons.

A test of a concrete type would be like `objType == GetObjectType(wxT("sizer"))`

src/model/database.cpp
```diff
@@ -301,9 +314,23 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
...
+			{
+				std::set<wxString> types;
+				types.insert(wxT("sizer"));
+				types.insert(wxT("gbsizer"));
+				count = CountChildrenWithTypes(parent, types);
```
sodevel: wxFB currently requires at least C++11 (actually even C++17 is already used in one place) so you can use these language features to simplify this into `count = CountChildrenWithSameType(parent, { GetObjectType(wxT("sizer")), GetObjectType(wxT("gbsizer")) });`

src/model/database.cpp
```diff
@@ -301,9 +314,23 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
...
+			}
+			else 
+			{
+				count = CountChildrenWithSameType(parent, objType);
+			}
+
```
sodevel: Another whitespace error

src/model/database.cpp
```diff
@@ -301,9 +314,23 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
...
+			}
+			else 
+			{
+				count = CountChildrenWithSameType(parent, objType);
+			}
+			
 
+			if (max > 0 && count >= max)
+				create = false;
+
```
sodevel: Another whitespace error

src/model/database.cpp
```diff
@@ -257,7 +270,7 @@ PObjectBase ObjectDatabase::CreateObject( std::string classname, PObjectBase par
 	}
 
 	PObjectType objType = objInfo->GetObjectType();
-
+
```
sodevel: Whitespace error

src/model/database.cpp
```diff
@@ -228,6 +228,19 @@ int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent,PObjectType typ
 	return count;
 }
 
+int ObjectDatabase::CountChildrenWithTypes(PObjectBase parent,const std::set<wxString> &types)
```
sodevel: Missing whitespace in front of the comma, attach the reference to the type please

src/model/database.cpp
```diff
@@ -228,6 +228,19 @@ int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent,PObjectType typ
...
+{
+	unsigned int count = 0;
+	unsigned int numChildren = parent->GetChildCount();
+	for (unsigned int i=0; i < numChildren ; i++)
```
sodevel: Missing whitespace around the equals sign, extra whitespace in front of the second semicolon.

Please don't use post-increment, always use pre-increment unless the use of post-increment is required

src/model/database.cpp
```diff
@@ -228,6 +228,19 @@ int ObjectDatabase::CountChildrenWithSameType(PObjectBase parent,PObjectType typ
...
+	{
+		if (types.find(parent->GetChild(i)->GetObjectInfo()->GetObjectType()->GetName()) != types.end())
+			count++;
```
sodevel: Unnecessary post-increment

src/model/database.cpp
Kvaz1r: It's really great since wxWidgets itself still requires C++98.

Kvaz1r:
Yes, I will fix all of your remarks on the weekends. As for whitespace, it's no more tabs - four space or something else?

sodevel:
In source files and code template files as indentation currently one tab is used. Only in xml files two spaces are used.

Issue #217, opened 2015-02-22
Generated XRC use old namespace


wxfbint:
wxFormBuilder creates XRC code with the old namespace http://www.wxwindows.org/wxxrc instead of the new one http://www.wxwidgets.org/wxxrc .

See http://docs.wxwidgets.org/3.0.0/overview_xrcformat.html .

I’m using wxFormBuilder 3.5.1-RC1 (3.5.01-0ubuntu1~utopic1).

Reported by: bmh1980

PR #539, opened 2019-08-27
Update xrc namespace

Kvaz1r:
Fixes: #217


src/codegen/xrccg.cpp
```diff
@@ -46,7 +46,7 @@ bool XrcCodeGenerator::GenerateCode( PObjectBase project )
 	doc.LinkEndChild( &decl );
 
 	ticpp::Element element( "resource" );
-	element.SetAttribute( "xmlns", "http://www.wxwindows.org/wxxrc" );
+	element.SetAttribute( "xmlns", "http://www.wxwidgets.org/wxxrc" );
```
sodevel: New code shouldn't use that whitespace padding of parentheses you can find frequently in the code base, can you remove that please?

Issue #29, opened 2011-05-23
OnMove event support


wxfbint:
Support OnMove event for wxwindow which isn't supported yet 

Reported by: *anonymous

PR #545, opened 2019-10-01
Update OnMove event support

Kvaz1r:
Fix: #29


sodevel:
Aren't Move-Events only generated for Toplevel-Windows? At least in the documentation of the latest sources i can find only this information, what makes you think that these events also work for every other wxWindow?

Kvaz1r:
It [was fixed](https://github.com/wxWidgets/wxWidgets/commit/7a1bc568e5cb8d04dead3a1f53c3563ce41d4fef) in the documentation recently, but it also relates for 3.0 branch as well.

sodevel:
Ah thanks, looks like my 1.5 week old build is 1 week too old :smile:

But anyway i see another issue, you added a code template only for C++, this should cause errors already in the preview for the other languages. Can you add the required templates also for the other languages?

Kvaz1r:
At least I will try  ;-) Never used other languages so didn't know about it.

output/xml/default.xml
```diff
@@ -179,6 +179,7 @@
 		<event name="OnPaint" class="wxPaintEvent" help="Process a wxEVT_PAINT event." />
 		<event name="OnSize" class="wxSizeEvent" help="Process a wxEVT_SIZE event." />
 		<event name="OnUpdateUI" class="wxUpdateUIEvent" help="Process a wxEVT_UPDATE_UI event." />
+		<event name="OnMove" class="wxMoveEvent" help="Process a wxEVT_MOVE event, which is generated when a window is moved. Not generated in wxGTK" />
```
sodevel: wxFB displays the events in the order they appear in the xml file, OnSize and OnMove are kind of related and now separated by OnUpdateUI. Can you place it between OnSize and OnUpdateUI instead please and also update the order in the code templates accordingly?

output/xml/default.luacode
```diff
@@ -151,6 +151,9 @@
 
 		<!-- wxUpdateUIEvent -->
 		<template name="evt_connect_OnUpdateUI">#utbl$name:Connect( wx.wxEVT_UPDATE_UI, function(event)#nl --implements  #handler#nl #skip #nl end  )</template>
+		
+		<!-- wxSizeEvent -->
```
sodevel: Copy-Paste error

output/xml/default.xml
Kvaz1r: Fixed

PR #547, opened 2019-10-03
Added support for wxBitmapToggleButton

jmoraleda:
Added support for `wxBitmapToggleButton` (https://docs.wxwidgets.org/trunk/classwx_bitmap_toggle_button.html) in the _Additional_ tab of the _Component Palette_, next to the `wxToggleButton` component.

I added support for all languages that wxFormBuilder can generate templates for.


sodevel:
I am quite unsure what to do about this. Isn't the wxBitmapToggleButton superfluous like the wxBitmapButton? Since the wxButton rework of wxWidgets all buttons can display bitmaps, maybe the wxToggleButton and wxBitmapToggleButton implementation should be updated like the wxButton and wxBitmapButton one was before. These two buttons even share all properties of a regular wxButton minus the default ability and auth ability (which is currently missing in wxFB), i don't know if its worth to extract these base properties into a base class and use the inherits feature to use this in all four button types.

The second issue is the code style of the plugin code. I know this is just a copy-paste of the wxToggleButton code, but this code doesn't adhere to the current coding style and usually i don't accept this for new code. I might add a code review about these details later but i'm currently short on time. I think it's worth to think about the button refactoring above or at least use the inherits feature between both toggle buttons.

jmoraleda:
Hello @sodevel. Thank you for your comprehensive reply. I very much agree that removing code duplication is very desirable. A common base class for all four button types sounds like a good idea. It would make sense to refactor all four buttons at once. 

I am happy to help if it would be useful, but I don't know the form builder plugin framework (as you realized I just copy pasted from toggle button taking the bitmap info from bitmap button). In particular, does inheritance work only in the cpp code, or also in the xml? 

If you can point me to an example with adhering to the current code style of multiple similar components sharing a base class, I could give the refactoring a try.

sodevel:
I don't know all the details about how that inheritance feature works, it's been a while i worked on the code, but maybe i can give a small overview.

Actually it is reversed, there is no inheritance used on the c++ side, the plugin code, allmost every component is implemented independent.

On the XML side however it is different, there exists the `inherits` tag which pulls in all definitions from the inherited element. You can see the effect in the object properties grid on the right side, these inherited definitions end up in subgroups. It is possible to override inherited properties, most prominent example is the name.

This also affects the code templates, all inherited code templates get applied automatically. It is possible to override these as well but this has different effects which i don't know in detail and discovered more or less by try and error. For example the construction template totally overrides the parent one but the settings templates get merged.

As an example of how to use inheritance take a look at wxButton and wxBitmapButton, about code style i currently have no better advice as to look at the commits and take a look at the `.clang-format` and `.clang-tidy` files.

jmoraleda:
I just looked at this again. Is it possible to inherit in  xml in the `additional` plugin from the 'common` plugin or do I need to have the base and derived components be part of the same plugin? If so, should I move the toggle button and toggle bitmap button to the common plugin?

sodevel:
Apparently `inherits` works across xml files, at least it works with `default.xml`, all plugins inherit elements from that one. To not create dependencies between plugins i would prefer that if you need a common base for different plugins to store that inside `default.xml` like `wxTreeCtrlBase` currently is. 

It's not that great to store something that is not of such great relevance like the other elements of that file inside there but there is currently no other common file to store elements that are required by multiple plugins.

jmoraleda:
I refactored the code so that all buttons in inherit from a common class `wxAnyButton` in 'default.xml`. Please take a look at it when you have a chance and check if this is what you had in mind. Thank you.

sodevel:
Thanks for your work, this is looking good and what i had in mind. Despite some small issues the style-police found there are only two bigger issues, but take a look at the code review. It would be great if you could fix these so i can merge this easily.

output/plugins/additional/xml/additional.pythoncode
```diff
@@ -57,6 +57,10 @@ Python code generation written by
 		<template name="settings">#ifnotequal $value "0" @{ self.$name.SetValue( True ) @}</template>
 		<template name="evt_connect_OnToggleButton">self.$name.Bind( wx.EVT_TOGGLEBUTTON, #handler )</template>
 	</templates>
+
```
sodevel: Trailing whitespace

output/plugins/additional/xml/additional.xml
```diff
@@ -81,16 +81,23 @@ Written by
     <property name="tab_width" type="text" help="The number of spaces per tab character.">4</property>
   </objectinfo>
 
-  <objectinfo class="wxToggleButton" startgroup="1" icon="toggle_button.xpm" type="widget">
+
```
sodevel: Extra empty line

output/plugins/additional/xml/additional.xml
```diff
@@ -81,16 +81,23 @@ Written by
...
+  <objectinfo class="wxToggleButton" icon="toggle_button.xpm" type="widget">
```
sodevel: Any specifiy reason why this element shouldn't start a group anymore? I think it makes even more sense now that there is an additional toggle-button

output/plugins/additional/xml/additional.xml
```diff
@@ -81,16 +81,23 @@ Written by
...
     <inherits class="wxWindow"/>
     <inherits class="AUI"/>
     <inherits class="Validator"/>
     <property name="name" type="text">m_toggleBtn</property>
-    <property name="label" type="wxString_i18n">Toggle me!</property>
     <property name="value" type="bool">0</property>
     <event name="OnToggleButton" class="wxCommandEvent" help="Handles a wxEVT_TOGGLEBUTTON event."/>
   </objectinfo>
 
+  <objectinfo class="wxBitmapToggleButton" icon="bitmap_toggle_button.xpm" type="widget">
```
sodevel: The specified icon file does not exist, i think you had one before, maybe it got lost during force-push?

output/plugins/common/xml/common.cppcode
```diff
@@ -26,34 +26,7 @@ Written by
...
-			#ifnotnull $current
-			@{ #nl $name->SetBitmapCurrent( $current ); @}
-
-			#ifnotnull $position
-			@{ #nl $name->SetBitmapPosition( $position ); @}
-
-			#ifnotnull $margins
-			@{ #nl $name->SetBitmapMargins( $margins ); @}
-		</template>
+		<template name="settings"> #ifnotequal $default "0" @{ #nl $name->SetDefault(); @} </template>
```
sodevel: Extra whitespace around the tag content

output/plugins/common/xml/common.luacode
```diff
@@ -30,34 +30,7 @@ Lua code generation written by
...
-			#ifnotnull $current
-			@{ #nl #utbl$name:SetBitmapCurrent( $current ) @}
-
-			#ifnotnull $position
-			@{ #nl #utbl$name:SetBitmapPosition( $position ) @}
-
-			#ifnotnull $margins
-			@{ #nl #utbl$name:SetBitmapMargins( $margins ) @}
-		</template>
+		<template name="settings"> #ifnotequal $default "0" @{ #nl #utbl$name:SetDefault() @} </template>
```
sodevel: Extra whitespace around the tag content

output/plugins/common/xml/common.phpcode
```diff
@@ -30,34 +30,7 @@ PHP code generation written by
...
-			#ifnotnull $current
-			@{ #nl @$this->$name->SetBitmapCurrent( $current ); @}
-
-			#ifnotnull $position
-			@{ #nl @$this->$name->SetBitmapPosition( $position ); @}
-
-			#ifnotnull $margins
-			@{ #nl @$this->$name->SetBitmapMargins( $margins ); @}
-		</template>
+		<template name="settings"> #ifnotequal $default "0" @{ #nl @$this->$name->SetDefault(); @} </template>
```
sodevel: Extra whitespace around the tag content

output/plugins/common/xml/common.pythoncode
```diff
@@ -30,34 +30,7 @@ Python code generation written by
...
-			#ifnotnull $current
-			@{ #nl self.$name.SetBitmapCurrent( $current ) @}
-
-			#ifnotnull $position
-			@{ #nl self.$name.SetBitmapPosition( $position ) @}
-
-			#ifnotnull $margins
-			@{ #nl self.$name.SetBitmapMargins( $margins ) @}
-		</template>
+		<template name="settings"> #ifnotequal $default "0" @{ #nl self.$name.SetDefault() @} </template>
```
sodevel: Extra whitespace around the tag content

output/plugins/common/xml/common.xml
```diff
@@ -24,35 +24,21 @@ Written by
...
+      <option name="wxBU_NOTEXT"    help="Disables the display of the text label in the button even if it has one or its id is one of the standard stock ids with an associated label: without using this style a button which is only supposed to show a bitmap but uses a standard id would display a label too."/>
+      <option name="wxBORDER_NONE"  help="Creates a button without border. This is currently implemented in MSW, GTK2 and OSX/Cocoa and OSX/Carbon ports but in the latter only applies to buttons with bitmaps and using bitmap of one of the standard sizes only, namely 128*128, 48*48, 24*24 or 16*16. In all the other cases wxBORDER_NONE is ignored under OSX/Carbon (these restrictions don't exist in OSX/Cocoa however)."/>
```
sodevel: Don't align the help messages, this can cause editing avalanches

output/xml/default.xml
```diff
@@ -203,6 +203,24 @@
...
+    <property name="disabled" type="bitmap" help="Bitmap shown when the button is disabled."/>
+    <property name="pressed" type="bitmap" help="Bitmap shown when the button is pushed (e.g. while the user keeps the mouse button pressed on it)."/>
+    <property name="focus" type="bitmap" help="Bitmap shown when the button has keyboard focus but is not pressed."/>
+    <property name="current" type="bitmap" help="Bitmap shown when the mouse is over the button (but it is not pressed). Notice that if hover bitmap is not specified but the current platform UI uses hover images for the buttons (such as Windows XP or GTK+), then the focus bitmap is used for hover state as well. This makes it possible to set focus bitmap only to get reasonably good behaviour on all platforms."/>
+    <property name="position" type="option">
+      <option name=""/>
+      <option name="wxLEFT"   help="Positions the bitmap at the left"/>
+      <option name="wxRIGHT"  help="Positions the bitmap at the right"/>
+      <option name="wxTOP"    help="Positions the bitmap at the top"/>
+      <option name="wxBOTTOM" help="Positions the bitmap at the bottom"/>
```
sodevel: Don't align the help messages, this can cause editing avalanches

output/xml/default.phpcode
```diff
@@ -187,6 +187,19 @@
...
+			#ifnotnull $bitmap @{ #nl @$this->$name->SetBitmap( $bitmap ); @}
+			#ifnotnull $disabled @{ #nl @$this->$name->SetBitmapDisabled( $disabled ); @}
+			#ifnotnull $pressed @{ #nl @$this->$name->SetBitmapSelected( $pressed ); @}
+			#ifnotnull $focus @{ #nl @$this->$name->SetBitmapFocus( $focus ); @}
+			#ifnotnull $current @{ #nl @$this->$name->SetBitmapCurrent( $current ); @}
+			#ifnotnull $position @{ #nl @$this->$name->SetBitmapPosition( $position ); @}
+			#ifnotnull $margins @{ #nl @$this->$name->SetBitmapMargins( $margins ); @}
+		</template>
+	</templates>
+
```
sodevel: Trailing whitespace

output/xml/default.cppcode
```diff
@@ -281,6 +281,19 @@
 		<template name="evt_connect_OnTreeItemMenu">$name->Connect( wxEVT_COMMAND_TREE_ITEM_MENU, #handler, NULL, this );</template>
 	</templates>
 
+	<templates class="wxAnyButton">
+		<template name="settings">
+			#ifequal $markup "1" @{ $name->SetLabelMarkup( $label );  @}
```
sodevel: Extra whitespace before the closing @}

plugins/additional/additional.cpp
```diff
@@ -416,19 +416,54 @@ class HtmlWindowComponent : public ComponentBase
...
+		if (!obj->IsNull(_("position"))) {
+			button->SetBitmapPosition(
+			    static_cast<wxDirection>(obj->GetPropertyAsInteger(_("position"))));
+		}
+
+		if (!obj->IsNull(_("margins"))) {
+			button->SetBitmapMargins(obj->GetPropertyAsSize(_("margins")));
+		}
+
+		button->SetValue( ( obj->GetPropertyAsInteger(_("value")) != 0 ) );
```
sodevel: Extra padding around parentheses

plugins/additional/additional.cpp
```diff
@@ -416,19 +416,54 @@ class HtmlWindowComponent : public ComponentBase
...
+		wxString label = obj->GetPropertyAsString( _("label") );
+		wxToggleButton* button = new wxToggleButton((wxWindow *)parent, wxID_ANY,
+			label,
 			obj->GetPropertyAsPoint(_("pos")),
 			obj->GetPropertyAsSize(_("size")),
-			obj->GetPropertyAsInteger(_("window_style")));
+			obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
 
-		window->SetValue( ( obj->GetPropertyAsInteger(_("value")) != 0 ) );
-		window->Connect( wxEVT_COMMAND_TOGGLEBUTTON_CLICKED, wxCommandEventHandler( ToggleButtonComponent::OnToggle ), NULL, this );
```
sodevel: This removal breaks the updating of the value if the button gets clicked in the preview, i see no reason why this feature should be removed

plugins/additional/additional.cpp
```diff
@@ -443,27 +478,151 @@ class ToggleButtonComponent : public ComponentBase, public wxEvtHandler
...
+			obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
+
+#if wxCHECK_VERSION( 2, 9, 2 )
+		// To stay in sync what the generator templates do apply the markup label here as well
+		if (obj->GetPropertyAsInteger(_("markup")) != 0) {
+			button->SetLabelMarkup(obj->GetPropertyAsString(_("label")));
+		}
+#endif
+
+		if ( !obj->IsNull( _("disabled") ) )
```
sodevel: Extra padding around parentheses

plugins/additional/additional.cpp
```diff
@@ -443,27 +478,151 @@ class ToggleButtonComponent : public ComponentBase, public wxEvtHandler
...
+		{
+			button->SetBitmapDisabled( obj->GetPropertyAsBitmap( _("disabled") ) );
+		}
+
+		if (!obj->IsNull(_("pressed"))) {
+			button->SetBitmapPressed(obj->GetPropertyAsBitmap(_("pressed")));
+		}
+
+		if ( !obj->IsNull( _("focus") ) )
```
sodevel: Extra padding around parentheses

plugins/additional/additional.cpp
```diff
@@ -443,27 +478,151 @@ class ToggleButtonComponent : public ComponentBase, public wxEvtHandler
...
+		if (!obj->IsNull(_("margins"))) {
+			button->SetBitmapMargins(obj->GetPropertyAsSize(_("margins")));
+		}
+
+		button->SetValue( ( obj->GetPropertyAsInteger(_("value")) != 0 ) );
+
+		return button;
+	}
+
+	void OnToggle( wxCommandEvent& event )
```
sodevel: Extra padding around parentheses

plugins/additional/additional.cpp
```diff
@@ -443,27 +478,151 @@ class ToggleButtonComponent : public ComponentBase, public wxEvtHandler
...
+			button->SetBitmapPosition(
+			    static_cast<wxDirection>(obj->GetPropertyAsInteger(_("position"))));
+		}
+
+		if (!obj->IsNull(_("margins"))) {
+			button->SetBitmapMargins(obj->GetPropertyAsSize(_("margins")));
+		}
+
+		button->SetValue( ( obj->GetPropertyAsInteger(_("value")) != 0 ) );
+
```
sodevel: Here too the connect of the wxEVT_COMMAND_TOGGLEBUTTON_CLICKED is missing so that button clicks don't update the value in the model

jmoraleda:
Thank you!

PR #550, opened 2019-10-10
Reformat XML files and code templates

sodevel:
I am placing this PR to start a discussion and collect some opinions.

The current wxFB codebase is in a bad shape which makes it difficult to read and work with. There is a mix of different coding styles and especially formatting and whitespace styles. Even if you work regulary on the code it is kind of difficult to remember how to format things. It is also hard to use automatic code formatters, they tend to format the surrounding code or even worse the whole file which introduces lots of formatting changes that blend with functional changes. I have spend quite too much time on reverting automatic formatting so the commits contain only functional changes.

The current policy is to write new code according to the current formatting style, while this helps to update the formatting it produces hard to read code if the surrounding code is a lot different, not to mention how difficult it can be to align the code in such an environment.

I am not sure how to deal with that situation properly. In the company i work for we solved this by simply reformatting the whole codebase in one step and use the new style from that time on. But there we don't have that kind of "social component" how many lines someone changed like here on GitHub and also do not need the `blame` function.

In the past, when i updated the XML files and Code-Templates, i also updated their formatting to a uniform style. @jhasse took the time to carefully remove these formatting changes before he merged them, i have kept them around and updated them which was quite tedious. But in my opinion it is much easier to work with these so this PR contains these changes so you can get an own impression.

This PR contains pure formatting changes, no functional changes. Mostly whitespace conversion and indentation fixes, some styles changes to use an uniform style. These changes are not so drastic because most files already had a very similar style, this is a bit different from the situation of the c++ source code. I would merge this PR, maybe using a bot account or something so that all the "fame" gets concentrated there, but i would like to hear what others think about this.


Kvaz1r:
I don't think that many C++ developers really cares about amount of added/removed lines on Github so my voice for merging without any doubts ;-)
Concerted code style would be really useful for those who only started read codebase(like me).

jhasse:
I do care, but I'm no longer working on wxFormBuilder. So go for it :)

Issue #533, opened 2019-07-29
Install from source doesn't work for Windows(can't run after building). 


Kvaz1r:
.exe start running only when I manually transferred .dll files from old version.  

I see couple similar issues but they already closed.

sodevel:
How do you build on Windows, with Meson or PreMake? What DLL files you need to copy?

At least PreMake does not take care of external DLLs, this is the wxWidgets DLL and the ones your compiler needs. You have to make sure that these are available on the search path, easiest way is to copy them in the output folder. This is not an error but kind of a limitation of the build system.

Kvaz1r:
Via MSYS commands from Readme.
I don't know what exactly .dll was needed so just copied all of them (wxwidget , other dependency and standard (libstdc++-6, libgcc_s_dw2-1)).

sodevel:
This is the expected behavior. You can use a dependency checker to figure out the exact dependencies but beside the single wxWidgets DLL there are only the compiler DLL's left, like in your case the C++ runtime DLL and DWARF-2 MinGW DLL.

Kvaz1r:
Yes, it's not the point. I think it would be better write about it in Readme (especially if it's expected), because it's common issue and people back into it again and again.

PR #552, opened 2019-10-14
Update info about windows install from source

Kvaz1r:
I still think it worth to mention about needing copy dlls when install with MSYS so I propose such changes. 

Fixes: #533

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ sed 's!-lbfd!-lbfd -lz!g' *.make -i
 make config=release
 ```
 
-Run:
+Run (you may also need copy several .dll into output dir from mingw32\bin):
 
 ```sh
 cd ../../../output/


sodevel:
The reason in your instructions is not quite correct, these DLL's don't need to be in the execution directory, they need to be on the search path. Granted on Windows the easiest way to get them into the search path is to put them into the directory of the executable.

The instructions itself are quite vague, i don't know how much this helps the average user. But i also don't know how it would help if you mention that the compiler runtime and c++ runtime with its dependencies are required.

Maybe you can find a better wording for the issue?

Kvaz1r:
> The reason in your instructions is not quite correct, these DLL's don't need to be in the execution directory, they need to be on the search path.

good point. 

> The instructions itself are quite vague, i don't know how much this helps the average user. 

Maybe, and I'm not quite sure that it really worth it, that is why I asked about it in issue. If you sure that it's needless - let's close it.

Kvaz1r:
After re-thinking I decided to close it.

PR #553, opened 2019-10-16
Bug with subclass in components in python

JacoboSpain:
Problem with subclasses in some components in python.

I am using the latest version of the repository in windows:

When the subclass is configured in some components this is not reflected in the generated .py. as seen in the image.
For example to inherit the TreeListCtrl component of SubCLass_Name:
imagen

In other components it is done correctly: 

imagen

I have reviewed the xml, to correctly generate those components by replacing static assignments:

Original:
imagen

Modified:

imagen

They work right 

imagen

Thanks

Sorry my english is very bad.

### DIFF
--- a/output/plugins/additional/xml/additional.pythoncode
+++ b/output/plugins/additional/xml/additional.pythoncode
@@ -238,7 +238,7 @@ Python code generation written by
 
   <templates class="wxMediaCtrl">
     <template name="construction">
-		self.$name = wx.media.MediaCtrl( #wxparent $name, $id, wx.EmptyString, $pos, $size)
+		self.$name = #class( #wxparent $name, $id, wx.EmptyString, $pos, $size)
 		#ifnotnull $file
 		@{ #nl self.$name.Load( $file ) @}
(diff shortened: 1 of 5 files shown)


sodevel:
wxFB uses the class names of the original wxWidgets library, the wxPython language binding tries to use the same names by stripping off the wx prefix and use it as a module. The code generator can adjust the class names automatically, but certain classes in wxPython use different modules so the automatic fails.

For that reason currently there is a hardcoded list of classes and other symbols inside the python code generator class that provides special translations for such elements. All changes you made break the code generation because all of the changed elements, except wxMenu, cannot be used with the automatic adjustment but require a special translation. For these elements that special translation is not done inside the code generator class but hardcoded into the code templates.

I'm sorry but i cannot merge this. There is currently no other solution than hardcode to solve this problem and that solution is not ideal. If you really need the subclass feature for these elements you have to add the translation rules to the code generator class so that code generation does not break. That i can merge but should be replaced in the future with a better solution.

sodevel:
Closing stale PR

PR #557, opened 2019-11-13
Fix python code generation for image wrapper function (#496)

ANoDE85:
The feature I implemented in January is currently broken on master.

My initial implementation tried to be smart and only set the m_imagePathWrapperFunctionName member only when absolutely necessary. This proved to be fatal, since maintainers would have to keep track of when to call SetImagePathWrapperFunctionName on the parser (maybe even I already forgot a few places).

I've changed the implementation now so that the PythonTemplateParser constructor receives a new parameter that must be filled by instantiators. This way, the value is passed on to all instances correctly.

### DIFF
--- a/src/codegen/pythoncg.cpp
+++ b/src/codegen/pythoncg.cpp
@@ -40,12 +40,13 @@
 #include <wx/filename.h>
 #include <wx/tokenzr.h>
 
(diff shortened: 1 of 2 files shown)


sodevel:
Although the coding style doesn't adhere to the current policy it follows the surrounding style so i will accept this and instead question if the current policy should be kept. You know more about Python than i do so i can't judge the functionality but changes are looking good.

Issue #562, opened 2019-11-26
Storing a pointer into a long is preventing building 64-bit version.


Randalphwa:
### Description:
The primary obstacle to building **wxFormBuilder** as a 64-bit app is that **AuiToolBar** calls `wxAuiToolBarItem::SetUserData()` and `wxAuiToolBarItem::GetUserData()` for storing and retrieving a **wxObject*** pointer. Unfortunately, **wxAuiToolBarItem** uses a **long** rather than **ptrdiff_t** for storing this data, which means it is only viable for 32-bit pointers.

The proposed solution is to use an **unordered_map** to store the item number as a key, and the pointer as the value. This can be added to the **AuiToolBar** class with methods for setting and retrieving the pointer based on the item number (which gets stored instead of the pointer in Set/GetUserData).

PR #563, opened 2019-11-26
Stores wxObject pointer in a void* instead of a long.

Randalphwa:
Fixes # 562

The only significant issue preventing building **wxFormBuilder** and all plugins as a 64-bit target is issue #562 which refers to storing a pointer into a long.

On my fork I am now able to build and run a 64-bit version of **wxFormBuilder** and all plugin dlls. It does require an additional change to **ImportComponentLibrary** to load the correct version of the dll, however that's going to be dependent on build script changes, so I'm not including that change in this PR. In addition, just because it builds and runs, doesn't mean it's stable, so I want to spend a lot more time working with the 64-bit version before suggesting adding it as an optional build platform.

Note that I did not wrap the changes in bit-depth conditionals. While the change isn't strictly necessary for 32-bit builds, I don't think it will have any impact on performance. Not using conditionals in this case makes for easier to read code, and of course allows anyone else to build a 64-bit version.

### DIFF
--- a/plugins/common/common.cpp
+++ b/plugins/common/common.cpp
@@ -23,6 +23,8 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include  <unordered_map>
+
 #include <plugin.h>
 #include <ticpp.h>
 #include <xrcconv.h>
@@ -100,6 +102,15 @@ public:
 		m_manager = manager;
 	}
 
+	void SetPtrData(int index, void* ptr)
+	{
+		m_aObjects[index] = ptr;
+	}
+	void* GetPtrData(int index)
+	{
+		return m_aObjects[index];
+	}
+
 protected:
 	IManager* m_manager;
 
@@ -109,6 +120,9 @@ protected:
 	void OnTool( wxCommandEvent& event );
 
 	DECLARE_EVENT_TABLE()
+
+private:
+	std::unordered_map<int, void*> m_aObjects;
 };
 
 BEGIN_EVENT_TABLE( AuiToolBar, wxAuiToolBar )
@@ -1261,7 +1275,7 @@ void AuiToolBar::OnDropDownMenu( wxAuiToolBarEvent& event )
 
 		if ( item && item->HasDropDown() )
 		{
-			wxObject* wxobject = (wxObject*) item->GetUserData();
+			wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
 
 			if ( NULL != wxobject )
 			{
@@ -1298,11 +1312,15 @@ void AuiToolBar::OnTool( wxCommandEvent& event )
 		return;
 	}
 
-	wxObject* wxobject = (wxObject*) tb->FindTool( event.GetId() )->GetUserData();
-
-	if ( NULL != wxobject )
+	wxAuiToolBarItem* item = tb->FindTool(event.GetId());
+	if (item)
 	{
-		m_manager->SelectObject( wxobject );
+		wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
+
+		if (NULL != wxobject)
+		{
+			m_manager->SelectObject(wxobject);
+		}
 	}
 }
 
@@ -1354,7 +1372,9 @@ public:
 								child
 							);
 				wxAuiToolBarItem* itm = tb->FindToolByIndex( i );
-				itm->SetUserData( (long) child );
+				wxASSERT(itm);
+				itm->SetUserData(i);
+				tb->SetPtrData(i, child);
 				if ( childObj->GetPropertyAsInteger(_("context_menu") ) == 1 && !itm->HasDropDown() )
 					tb->SetToolDropDown( itm->GetId(), true );
 				else if ( childObj->GetPropertyAsInteger(_("context_menu") ) == 0 && itm->HasDropDown() )
--- a/plugins/common/common.cpp
+++ b/plugins/common/common.cpp
@@ -102,11 +102,11 @@ public:
 		m_manager = manager;
 	}
 
-	void SetPtrData(int index, void* ptr)
+	void SetObject(int index, wxObject* pObject)
 	{
-		m_aObjects[index] = ptr;
+		m_aObjects[index] = pObject;
 	}
-	void* GetPtrData(int index)
+	wxObject* GetObject(int index)
 	{
 		return m_aObjects[index];
 	}
@@ -122,7 +122,7 @@ protected:
 	DECLARE_EVENT_TABLE()
 
 private:
-	std::unordered_map<int, void*> m_aObjects;
+	std::unordered_map<int, wxObject*> m_aObjects;
 };
 
 BEGIN_EVENT_TABLE( AuiToolBar, wxAuiToolBar )
@@ -1275,9 +1275,9 @@ void AuiToolBar::OnDropDownMenu( wxAuiToolBarEvent& event )
 
 		if ( item && item->HasDropDown() )
 		{
-			wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
+			wxObject* wxobject = GetObject(item->GetUserData());
 
-			if ( NULL != wxobject )
+			if (wxobject)
 			{
 				m_manager->SelectObject( wxobject );
 			}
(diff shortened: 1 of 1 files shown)


sodevel:
I am going to answer here and not in the Issue to focus the discussion in one place.

That is an interesting discovery you made there however i don't agree with your solution. Multiple wxWidgets elements offer the possibility of storing arbitrary user data but unfortunately they don't use the same approach everywhere. I think instead of fixing the issue on our side by implementing another level of indirection it would be the better option to fix the issue in wxWidgets itself. The wxDataViewListStore uses a wxUIntPtr to store the user data, i think it would be a viable solution in this case as well (or the wxIntPtr to keep it signed like it is now).

Randalphwa:
I certainly agree that the ideal approach is for the underlying library to change it's code to use platform-independent storage size for arbitrary data. However, that would mean it would be impossible to build a 64-bit version of wxFormBuilder until/if that library was changed, and it would further require building only against that library (or a later version).

Another way of handling this would be to wrap the changes in 64-bit conditionals -- it does, after all, only affect 64-bit builds. Then later on if the wxAui library gets updated, the conditionals could be changed to add that particular library version (i.e., don't use our indirection if building against a library where the UserData we need has been changed to be platform independent).

The reason I didn't suggest that in the first place is it makes the code a bit harder to read. And of course since currently I seem to be the only one desiring a 64-bit build, I could simply leave the change in my fork for possible future inclusion should the request for a 64-bit version become more common.

sodevel:
I agree that #ifdefs should be used sparingly and also see the issue regarding the library update. Personally i only build on Windows and use some recent master snapshot but this is currently not an option for the other platforms, there we have to use what the distribution offers.

I just saw this change is affecting only one component of one plugin, i think this doesn't add too much complexity and can be added. I have only a few remarks in the following review.

plugins/common/common.cpp
```diff
@@ -109,6 +120,9 @@ class AuiToolBar : public wxAuiToolBar
 	void OnTool( wxCommandEvent& event );
 
 	DECLARE_EVENT_TABLE()
+
+private:
+	std::unordered_map<int, void*> m_aObjects;
```
sodevel: This element is used only locally and seems to store only `wxObject*` elements, i would prefer to use this as data type so the casts can be avoided

plugins/common/common.cpp
```diff
@@ -1298,11 +1312,15 @@ void AuiToolBar::OnTool( wxCommandEvent& event )
...
-	wxObject* wxobject = (wxObject*) tb->FindTool( event.GetId() )->GetUserData();
-
-	if ( NULL != wxobject )
+	wxAuiToolBarItem* item = tb->FindTool(event.GetId());
+	if (item)
 	{
-		m_manager->SelectObject( wxobject );
+		wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
+
+		if (NULL != wxobject)
```
sodevel: Modern-C++: Don't use `NULL` for pointer values and don't compare pointers explicit to `nullptr`, a simple implicit test `if (wxobject)` is preferred

plugins/common/common.cpp
```diff
@@ -1298,11 +1312,15 @@ void AuiToolBar::OnTool( wxCommandEvent& event )
...
 	}
 
-	wxObject* wxobject = (wxObject*) tb->FindTool( event.GetId() )->GetUserData();
-
-	if ( NULL != wxobject )
+	wxAuiToolBarItem* item = tb->FindTool(event.GetId());
+	if (item)
 	{
-		m_manager->SelectObject( wxobject );
+		wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
```
sodevel: Modern-C++: Please no c-style casts. I'm quite a fan of `auto`, especially if the type is already used explicit on the right side, so this could be changed to `auto* wxobject = static_cast<wxObject*>(GetPtrData(item->GetUserData()));`. However this isn't necessary anymore if the lookup map returns `wxObject*`.

plugins/common/common.cpp
```diff
@@ -1261,7 +1275,7 @@ void AuiToolBar::OnDropDownMenu( wxAuiToolBarEvent& event )
 
 		if ( item && item->HasDropDown() )
 		{
-			wxObject* wxobject = (wxObject*) item->GetUserData();
+			wxObject* wxobject = (wxObject*) GetPtrData(item->GetUserData());
```
sodevel: Modern-C++, see below

plugins/common/common.cpp
```diff
@@ -1298,11 +1312,15 @@ void AuiToolBar::OnTool( wxCommandEvent& event )
 		return;
 	}
 
-	wxObject* wxobject = (wxObject*) tb->FindTool( event.GetId() )->GetUserData();
-
-	if ( NULL != wxobject )
+	wxAuiToolBarItem* item = tb->FindTool(event.GetId());
```
sodevel: Modern-C++, see below

Randalphwa:
I've been trying to match the coding practice that I'm seeing in other areas of the code, even if it's significantly different then how I would normally write the code. So for example, while I would never write `if (NULL == ptr)` I've seen that being used in other spots, so I did the same to match what I thought was common coding practice for this project. I realize a lot of people have worked on the code and both formatting style and coding methods vary quite a bit. My question is, is there a code module that you would consider to be a fairly good example of the type of coding practices/formatting you would like to see that I could refer to?

The .clang-format for the project only has a few settings. I tried running a few files through it and clang made massive changes, so it clearly can not be used file-wide. I took my own .clang-format, replaced the options that were different and use that to format any code that I work on, but I'm wonder if there is also a source file that contains the formatting you wish all the code was formatted as? If so, I can then use that as a template to make any changes (if needed) to my fork version of .clang-format to cover _only_ any new/changed code that I write?

Finally, the first PR check failed with the error "clang: resource not found". Since I was only checking in a source file change, I assume that's a scripting issue on the server end -- but is there anything I can do on my end to prevent that PR check error?

sodevel:
The .clang-format together with the policy that every new line of code should follow the style imposed by it as good as possible without touching any surrounding code was introduced by the now kind of fully retired maintainer. I did disagree with certain style elements (especially tab's and braces on next line) but followed the style. Now that i am kind of the sole maintainer i wanted to get some order into the formatting and style so it is easier to use automatic formatting but it turns out this is not so easy.

I applied my formatting changes to the xml files and code templates i did years ago and maintained in our own private fork recently (although im not sure if i should regret not having replaced the tab's in the templates with spaces). I tried every setting of clang-format to format the source but couldn't get a satisfactory result. Biggest problem i have with clang-format is how it splits long lines into multiple lines, i don't want the parameters on column 70 to drop down one by one over 10 lines, i want them to move on the next line and get indented by one level. And then the closing parenthesis on the next line to close that block. Since two years somone is working on a patch to get that closing parenthesis on the next line, hopefully that will be finished soon so i can give it another try.

Best advice i can give on formatting is use common sense and try to blend into the surrounding code. One thing you should not use anymore is that whitespace padding of parentheses. Hopefully somewhere in the near future i find a solution how to deal with current code base and put some auto formatting configuration in place so that there is a more formal definition available and that most importantly can be applied automatically. As a hint what to do today i can only say take a look at the recent commits.

About coding style, Modern C++ is welcome, but because compiler support for C++17 is quite low currently try not to use much of that, i think it is currently only enabled to support some newer attributes.

The CI was also realized by the former maintainer and im not very familiar with it. It seems that Drone is the only one that cannot be configured through the repository, it broke only recently and i hope it is an error on their side and fixes itself like it did with Travis some time ago. If not i think i will simply disable it, i think it was only used to run some clang analyzers over the code anyway. You can just ignore this error for now.

Randalphwa:
I changed the function names to GetObject and SetObject since they are now specifically storing/retrieving a wxObject* pointer.

In this particular case, I'm reluctant to use **auto** because neither the variable name or the function name indicates that it is a pointer being returned, not a reference or copy. I realize that currently the variable is not being dereferenced, but I would expect someone reading that code to do something like `wxobject.IsKindOf(...)` which of course fails because it should be `wxobject->IsKindOf(...)`. It probably comes from 30+ years of using Hungarian notation -- it just looks wrong to use **->** on a variable that doesn't start with **p**... :-)

Randalphwa:
AppVeyor is now failing with the complaint that InnoSetup is not installed. Hopefully like the drone check this will get fixed on the server side.

sodevel:
For pointers we are using an explicit `auto*` for better readability, of the hungarian notation we only use the visibility part for globals/statics/members. Although i do prefer `auto`, especially for storing results of method calls, to make refactoring easier it is currently not a requirement to use it.

Changes are looking good, thanks for your work.

PR #566, opened 2019-11-30
Enables/Disables Save button based on modified state of project file.

Randalphwa:
## Description

This PR enables/disables the Save menu item on the File menu, as well as the Save button on the toolbar based on whether the current project has been modified. It's added to the UpdateFrame() function that handles all the other enable/disable of UI elements.

### Notes

I'm so used to looking at the Save button to confirm that my current file has been saved, that when I first used **wxFB** I clicked the button multiple times wondering why my project wasn't being saved. Since **wxFB** already does a bunch of enabling/disabling it seemed to make sense to add the code that would bring this in line with common practice (at least on Windows -- I don't know if it's common on OS/X or Unix).

### DIFF
--- a/src/rad/mainframe.cpp
+++ b/src/rad/mainframe.cpp
@@ -965,8 +965,13 @@ void MainFrame::UpdateFrame()
 	GetStatusBar()->SetStatusText( filename, STATUS_FIELD_PATH );
 
 	// Enable/Disable toolbar and menu entries
-	wxMenu* menuEdit = GetMenuBar()->GetMenu( GetMenuBar()->FindMenu( wxT( "Edit" ) ) );
 	wxToolBar* toolbar = GetToolBar();
+
+	wxMenu* menuFile = GetMenuBar()->GetMenu(GetMenuBar()->FindMenu(_("File")));
+	menuFile->Enable(ID_SAVE_PRJ, AppData()->IsModified());
+	toolbar->EnableTool(ID_SAVE_PRJ, AppData()->IsModified());
+
+	wxMenu* menuEdit = GetMenuBar()->GetMenu( GetMenuBar()->FindMenu( wxT( "Edit" ) ) );
 
 	bool redo = AppData()->CanRedo();
 	menuEdit->Enable( ID_REDO, redo );


sodevel:
This is a nice idea, however currently it only disables the tool but not the menu entry as well. Can you disable the menu entry as well please for a more consistent behavior?

Randalphwa:
Something I noticed while adding the enable/disable for the Save menu item is that there are three places in mainframe.cpp where wxT() is used to access a menu rather than _(). I'd be happy to do a PR that fixes those, however I was also thinking that later on I'd like to run xgettext, create a dummy version of the .po file that changes all the text in an identifiable way, and use that to check all the UI that's at least obviously visible to track down and fix the places where wxT() was incorrectly used, and at least some of the cases where _() is incorrectly used. I also suspect localizing would create a fair number of crashes since at least in mainframe.cpp there are no checks to confirm a menu pointer was actually retrieved -- which it won't be if the name is localized and wxT() accesses it.

src/rad/mainframe.cpp
```diff
@@ -968,6 +968,11 @@ void MainFrame::UpdateFrame()
 	wxMenu* menuEdit = GetMenuBar()->GetMenu( GetMenuBar()->FindMenu( wxT( "Edit" ) ) );
 	wxToolBar* toolbar = GetToolBar();
 
+	toolbar->EnableTool(ID_SAVE_PRJ, AppData()->IsModified());
+	wxMenu* pmenuFile = GetMenuBar()->GetMenu(GetMenuBar()->FindMenu(_("File")));
```
sodevel: No type encoding in the name, we only use m_, s_ and maybe g_ from the hungarian notation, remove the p here please

src/rad/mainframe.cpp
```diff
@@ -968,6 +968,11 @@ void MainFrame::UpdateFrame()
...
+	if (pmenuFile)
```
sodevel: The file menu should be always there, at most i would use an assertion here to track down the _() and wxT() mismatches, but otherwise a crash if the menu is missing is ok

sodevel:
As i said before, there are a lot of places where these macros are used wrong so you might find some places where they are mixed wrong as well. It would be nice to fix these issues but i'm not sure if your propsed approach is suitable.

I think there are more places invisible in the UI that are affected, maybe it would be better to extract all types of strings and check their context and decide on that what type of macro is required.

src/rad/mainframe.cpp
```diff
@@ -968,6 +968,10 @@ void MainFrame::UpdateFrame()
 	wxMenu* menuEdit = GetMenuBar()->GetMenu( GetMenuBar()->FindMenu( wxT( "Edit" ) ) );
 	wxToolBar* toolbar = GetToolBar();
 
+	toolbar->EnableTool(ID_SAVE_PRJ, AppData()->IsModified());
+	wxMenu* menuFile = GetMenuBar()->GetMenu(GetMenuBar()->FindMenu(_("File")));
+	menuFile->Enable(ID_SAVE_PRJ, AppData()->IsModified());
```
sodevel: I don't want to nitpick too much but looking at the whole section of that file i can see that there is a certain order of statements and this order is violated here.

It would look more uniform if the retrieval of the file menu would be right above the retrieval of the edit menu and if you first enable the menu entry and then the toolbar entry. Can you reorder the statements please?

Randalphwa: Since the rest of that code block is using **toolbar**, I moved that just under the comment that starts the section, and then moved the File enabling code above the edit menu section so that the code flows in a more expected way.

Randalphwa: BTW, I just want to mention that I appreciate the feedback you are giving on these PRs. I'm new to the project, and your comments will help a lot in my getting a better understanding of what you're looking for in coding changes. I'd rather get a thorough handle on the preferred coding styles and practices on these minor PRs before I dive into anything more complex.

sodevel: These changes work for me. Now there is just one final step left i have to ask you for. All these fixup commits just clutter up the history, can you squash all these commits into one and force-push them? I could squash them on merge but i'd like to keep the branch visible in the history.

You are doing quite a lot of commits so can i ask you to take a look at https://chris.beams.io/posts/git-commit/ and take it as a guidance for your commit messages? Especially use the imperative form for the things you changed but not for the explanations why (see your other outstanding PR). As an example, in this PR the message would turn from `Enables/Disables Save button based on modified state of project file.` into `Enable/Disable save action based on modified state of project file`

sodevel:
Looks like something didn't work correctly, now there are two commits with almost an identical message, seems like your rebase stopped one commit too early. Can you try again and squash these two commits into one please?

PR #573, opened 2019-12-04
Changes UI interaction with user when loading an old project file

### LINKED ISSUE #567: Proposal for changes to UI dealing with old project files
A while back I used **wxFB** to look at some project files in a repository I had cloned. I got a somewhat odd message about the project file being old (the message is missing a verb and has some unusual sentence construction), and saw the bit about the old file being overwritten. I assumed that meant I would overwrite it if I clicked Save rather than Save As, so I continued opening various project files. Then I noticed that GitHub Desktop was telling me I had modified every .fbp file I opened, but apparently nothing else had changed. **wxFB** hadn't told me where it was saving the backup files to, so I just checked the .fbp files out again to restore them. While doing that I did notice the .bak files, but they had an old file date so I assumed they were part of the repository. It wasn't until I was looking at the **wxFB** code that I realized where it saved the backup files to.

While looking at the **wxFB** code, the other thing I noticed is that there is no check to see if the backup file was successfully created. While it's highly unlikely that it would fail, that's small comfort to someone that doesn't have a backup of the original file.

I have some proposals that might reduce the chances of someone else inadvertently changing a project file and not knowing how to get it back:

- Change the wording to say where the old file will be copied to and indicating that the new project will be automatically saved.
- Confirm that the backup copy was created. If not, give the user the option of trying again (so they can fix network issues, disk space, unprotecting the file, etc.), skipping the backup, or just cancelling opening the project file.

Longer term once I've had a chance to become a lot more familiar with the code base, I'd like to look at the conversion results which are currently displayed in a really long line that's hard to read -- it'd be nice to make that easier to read. The bigger issue is why the project needs to be saved in the first place -- I suspect that's a coding requirement rather than a design choice, but I do think it would better if reasonably possible to not automatically overwrite the original file unless the user asks to.

Randalphwa:
Fixes #567 

## Description

This changes the message that is displayed when **wxFB** opens an older project file. It states the name of the backup file, and indicates that the new project file will be saved. It then confirms that the backup file was created, and if not, gives the user a choice of retry, skip, cancel.

To see the dialog that is displayed if the copy fails (this assumes that wxfbTest.fbp is still an old version):

- cd to wxfbTest
- copy wxfbTest.fbp to wxfbTest.fbp.bak
- write-protect wxfbTest.fbp.bak
- open wxfbTest.fbp in **wxFB**

Note: This adds a debug-only log message that displays old/new version -- it's there purely for debugging convenience.

### DIFF
--- a/src/rad/appdata.cpp
+++ b/src/rad/appdata.cpp
@@ -1326,11 +1326,16 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
 			}
 
 			wxMessageBox(
-			    _("This project is using an older file format. It will be converted in memory while loading it.\n\n"
-			      "Caution! If you save the project, it will be overwritten with the newer file format."),
+			    _("This project file is using an older file format, it will be updated during loading.\n\n"
+			      "WARNING: Saving the project will update the format of the project file on disk!"),
 			    _("Older file format"));
 
-			if (!ConvertProject(doc, file, fbpVerMajor, fbpVerMinor))
+			if (ConvertProject(doc, file, fbpVerMajor, fbpVerMinor))
+			{
+				// Document has changed -- reacquire the root node
+				root = doc.FirstChildElement();
+			}
+			else
 			{
 				wxLogError(wxT("Unable to convert project"));
 				return false;
@@ -1377,7 +1382,6 @@ bool ApplicationData::ConvertProject(ticpp::Document& doc, const wxString& path,
 {
 	try
 	{
-		ticpp::Document doc;
 		XMLUtils::LoadXMLFile( doc, false, path );
 
 		ticpp::Element* root = doc.FirstChildElement();
--- a/src/rad/appdata.cpp
+++ b/src/rad/appdata.cpp
@@ -49,6 +49,7 @@
 #include <wx/fs_arc.h>
 #include <wx/fs_filter.h>
 #include <wx/fs_mem.h>
+#include <wx/richmsgdlg.h>
 #include <wx/tokenzr.h>
 
 using namespace TypeConv;
@@ -1323,24 +1324,15 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
 				wxLogError( wxT( "This project file is out of date.  Update your .fbp before using --generate" ) );
 				return false;
 			}
-			if ( wxYES == wxMessageBox( wxT( "This project file is not of the current version.\n" )
-			                            wxT( "Would you to attempt automatic conversion?\n\n" )
-			                            wxT( "NOTE: This will modify your project file on disk!" ), _( "Old Version" ), wxYES_NO ) )
-			{
-				// we make a backup of the project
-				::wxCopyFile( file, file + wxT( ".bak" ) );
 
-				if ( !ConvertProject( file, fbpVerMajor, fbpVerMinor ) )
-				{
-					wxLogError( wxT( "Unable to convert project" ) );
-					return false;
-				}
+			wxMessageBox(
+			    _("This project is using an older file format. It will be converted in memory while loading it.\n\n"
+			      "Caution! If you save the project, it will be overwritten with the newer file format."),
+			    _("Older file format"));
 
-				XMLUtils::LoadXMLFile( doc, false, file );
-				root = doc.FirstChildElement();
-			}
-			else
+			if (!ConvertProject(doc, file, fbpVerMajor, fbpVerMinor))
 			{
+				wxLogError(wxT("Unable to convert project"));
 				return false;
 			}
 		}
@@ -1363,7 +1355,8 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
 			PObjectBase old_proj = m_project;
 			m_project = proj;
 			m_selObj = m_project;
-			m_modFlag = false;
+			// Set the modification to true if the project was older and has been converted
+			m_modFlag = older;
 			m_cmdProc.Reset();
 			m_projectFile = file;
 			SetProjectPath( ::wxPathOnly( file ) );
@@ -1380,7 +1373,7 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
 	return true;
 }
 
-bool ApplicationData::ConvertProject( const wxString& path, int fileMajor, int fileMinor )
+bool ApplicationData::ConvertProject(ticpp::Document& doc, const wxString& path, int fileMajor, int fileMinor)
 {
 	try
 	{
@@ -1388,7 +1381,6 @@ bool ApplicationData::ConvertProject( const wxString& path, int fileMajor, int f
 		XMLUtils::LoadXMLFile( doc, false, path );
 
 		ticpp::Element* root = doc.FirstChildElement();
-
 		if ( root->Value() == std::string( "object" ) )
 		{
 			ConvertProjectProperties( root, path, fileMajor, fileMinor );
@@ -1428,8 +1420,6 @@ bool ApplicationData::ConvertProject( const wxString& path, int fileMajor, int f
 			fileVersion->SetAttribute( "major", m_fbpVerMajor );
 			fileVersion->SetAttribute( "minor", m_fbpVerMinor );
 		}
-
-		doc.SaveFile();
 	}
 	catch ( ticpp::Exception& ex )
 	{
--- a/src/rad/appdata.h
+++ b/src/rad/appdata.h
@@ -31,6 +31,7 @@
 
 namespace ticpp
 {
+class Document;
 
 class Node;
 
@@ -274,7 +275,8 @@ class ApplicationData
 		@param fileMinor The minor revision of the file
 		@return true if successful, false otherwise
 		*/
-		bool ConvertProject( const wxString& path, int fileMajor, int fileMinor );
+
+		bool ConvertProject(ticpp::Document& doc, const wxString& path, int fileMajor, int fileMinor);
 
 		/**
 		Recursive function used to convert the object tree in the project file to the latest version.


Randalphwa:
I'm at a loss here for dealing with the AppVeyor build failure. It's complaining that wxSysErrorMsgStr() is out of scope. I thought it might just be because <wx/log.h> wasn't #included in that module, but that didn't seem to help. It's suggesting I use wxSysErrorMsg() instead, but the wxWidgets documentation specifically says to use wxSysErrorMsgStr() instead of wxSysErrorMsg().

sodevel:
AppVeyor is using wxWidgets 3.0.4 and `wxSysErrorMsgStr()` is only available since wxWidgets 3.1.0

src/rad/appdata.cpp
```diff
@@ -1319,16 +1321,41 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
 			if( justGenerate ){
 				wxLogError( wxT( "This project file is out of date.  Update your .fbp before using --generate" ) );
 				return false;
 			}
-			if ( wxYES == wxMessageBox( wxT( "This project file is not of the current version.\n" )
-			                            wxT( "Would you to attempt automatic conversion?\n\n" )
-			                            wxT( "NOTE: This will modify your project file on disk!" ), _( "Old Version" ), wxYES_NO ) )
+
+			wxFileName fname(file);
+			wxString   strMsg, strOldBackupName(fname.GetFullName() + wxT(".bak"));
```
sodevel: This is an extremely tricky way to declare and initialize strOldBackupName, read: at first i had no idea what is going on here. We are doing C++ here, please don't use that C-style comma operator style here, declare that string on a separate line. Also don't align the consecutive declarations please. That direct initialization is also not really necessary, i think copy initialization looks more natural here.

src/rad/appdata.cpp
```diff
@@ -1319,16 +1321,41 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
+
+			if (wxMessageBox(strMsg.Format(_("This project file is an old version.\n\n"
+			                                 "Do you want to copy the old version to %s and save an updated version?"),
+			                               strOldBackupName),
```
sodevel: Old version sounds so negative, in my opinion not current version or something alike that sounds more friendly. Now that you are changing the message you could also make it more precise, it is not really the file itself that is not current but the file format.

We should use the same terms during the whole operation, here it is called copy, later save and the dialog title says backup. The term backup seems to be fitting best so i think the wording should be like create a backup or use backup as a verb.

src/rad/appdata.cpp
```diff
@@ -1319,16 +1321,41 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
-				// we make a backup of the project
-				::wxCopyFile( file, file + wxT( ".bak" ) );
+				{
+					wxLogNull logNo;  // Prevents wxCopyFile from displaying a wxWidgets error.
+					fname = file + wxT(".bak");
+
+					while (!::wxCopyFile(file, fname.GetFullPath()))
+					{
+						wxRichMessageDialog dlg(
+						    NULL, strMsg.Format(_("Unable to save old version to %s.\n\n"), strOldBackupName),
```
sodevel: Don't create a dialog without a parent, especially not a modal one. Don't call that static method through an object. Actually, why do you create that message in every loop? It should be constant and simply assigned outside the loop to strMsg.

src/rad/appdata.cpp
```diff
@@ -1319,16 +1321,41 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
+			                               strOldBackupName),
+			                 _("Old project file"), wxYES_NO) == wxYES)
 			{
-				// we make a backup of the project
-				::wxCopyFile( file, file + wxT( ".bak" ) );
+				{
+					wxLogNull logNo;  // Prevents wxCopyFile from displaying a wxWidgets error.
+					fname = file + wxT(".bak");
+
+					while (!::wxCopyFile(file, fname.GetFullPath()))
```
sodevel: Instead of constructing the string in every loop, you could create it once outside of the loop, it doesn't get changed. Actually i see no real purpose for the wxFileName object here, after all you construct the final path by simply string appending the second extension.

src/rad/appdata.cpp
```diff
@@ -1319,16 +1321,41 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
+						    _("Backup failed"), wxICON_WARNING | wxYES_NO | wxCANCEL);
+						dlg.SetYesNoCancelLabels(_("Retry"), _("Skip"), _("Cancel"));
+						dlg.ShowDetailedText(wxSysErrorMsg());
+						auto result = dlg.ShowModal();
+						if (result == wxID_NO)
+							break;  // User wants to skip backing up the file.
+						else if (result == wxID_CANCEL)
+							return false;  // User is cancelling converting the project.
+						else
+							continue;  // wxYES means retry.
```
sodevel: Please use braces also for one liners. I'm not a big fan of these inline comments, i would prefer them above the line they refer to. These are also no complete sentences so please to trailing point.

Randalphwa:
I'll make the requested coding style changes -- and apply it to future coding as well.

I'd actually like to rework this entire PR -- even tossing it and submitting a different one it if that makes more sense. From a user perspective, this PR still doesn't solve the two things I find annoying about the way wxFB handles and older file format:

- If the project file is under source control management, then there already is a backup. If the SCM isn't setup to exclude *.bak files, then it's going to want to track the project backup file, so you're stuck having to navigate to the directory and delete the file that wxFB just created.
- This still doesn't allow a user to preview project files on a repository without wxFB making changes to that repository -- and if you don't want to change it, then after you are done, you have to revert all the project files that wxFB updated.

Instead of writing a file that the user may have to then hunt down and delete, I'm thinking it would make more sense to ask the user if they want a backup, and if so, give them a Save As dialog so they aren't forced to create a backup in a repository (wxFB can default to the same directory as the original, but at least then the user can save it somewhere else, even changing the name if desired).

If the user doesn't want to modify the file, then essentially it becomes a read-only project file. At that point you either don't allow any changes at all unless a Save As is done, or you allow changes, but don't enable any Save actions until a Save As is done. That would bring it more in line with what a lot of editors and word processors do -- if you set them so they can edit a read-only document, then you have to use Save As to save the document.

The bottom line is that I want to avoid having to delete files or revert a project unless I specifically told wxFB to create a backup, or modify a project file.

sodevel:
I don't like the idea of that read-only mode, despite the difficulties to implement it (expanding and collapsing currently modifies the state so this would not be possible) i think it would be confusing to the user. What i do like is the idea to either skip the creation of the backup file or to store is somewhere else. This would simplify the use case of examining files under revision control that need to be converted, skip the backup and revert the changed file later.

To prevent the modification of the file at all it is required that the conversion could be done in memory. I don't know if this is currently possible or how difficult it is to imlpement that, but is this really necessary? What do other applications do in that case? Visual Studio requires rewriting of old project files and doesn't even make a backup.

Randalphwa:
The thing that's different with VS project files is that there's little reason to open one or more of them just to see how the project file was implemented. But if you see a dialog in an app and you wonder how they implemented the UI, you might want to open up the project file just to look at the layout. That's actually how I encountered it -- I was comparing the very different layout approach that another form designer takes, and I wanted to look at some existing projects just to see how wxFB was being used to create UI versus the other designer. 

With graphical applications like Photoshop, I don't recall there being any warning about loading an old format -- the conversion is done in memory, and if you save the file it's done in the new format unless you specifically tell it to save in compatibility mode with older versions. I don't remember if Word does that or not -- I think it just silently loads the older formats, and then it's up to the user if they want to save it in the new format (which is the default) or in some compatibility format. However both of these apps allow the user to open a file in the new format and save it (with a reduction in formatting or appearance) to an older format, so there's no permanent damage to saving a file in a new format.

I won't have a chance to really dive into this until sometime next week, but I wonder if this is a limitation of the tinyxml code -- perhaps it either won't allow creating an XML file in memory, or it will create in memory but won't allow you to then read it.

Randalphwa:
**wxFB**'s current process is to read the project file into a **ticpp::Document** class, and check the version. If it's an older format, it creates the document again, reads the project file into it, modifies the document in memory, saves it out to disk and then once again creates the document and reads the project file into it.

I can't see any reason in the code why the document needs to be created multiple times, or why it needs to be saved during the conversion. The conversion itself is done in the document class, the same as any other changes to the project. I've tried just doing the conversion in the already created document class and continuing without any additional reads or saves, and it's worked fine on several old project files I've loaded.

Currently, the --generate command will not work on an older project files, and instead tells the user they must update the project file first. A side effect of not forcing the save and just using the updated **ticpp::Document** class means that the --generate command works fine on older projects without requiring the user to load and resave the project file first.

That still leaves the question of what to tell/ask the user. If the user makes any changes to the project and saves it, it will be written in the current **wxFB** file format. Would it be enough just to present a message box to the user indicating that the older file format will be updated if the user makes any changes and saves the file?

sodevel:
Somehow this one got lost on my side so i took a quick look at your latest change to give you some feedback before the holidays.

In general i think the in-memory conversion is a good thing but i don't think it should be done in the --generate case. Isn't that supposed to run non-interactive? IIRC the conversion process can popup a dialog plus you popup a dialog as well, that breaks the non-interactive flow. Also the conversion can cause data loss so i think it's a better approach to deny non-interactive processing of outdated project files.

I think the notification about the conversion should be stronger emphasized and sound more like a warning. Maybe it should be displayed after the conversion so in case the conversion also causes an error message or even fails completely this message is displayed last or not at all.

Randalphwa:
There's already code in place that prevents the actual conversion warning/error message from displaying. However I agree that if there's any chance that updating the project file format could change the code generation, then wxFB should not be doing it. I've put back the original section that bails if `--generate` is being used. Note that while it appears that it is going to notify the user, it actually gives no indication of a problem occurring -- simply an error code when wxFB exits. It should be sending something to stderr but that would need to be changed elsewhere and would be a completely seperate issue.

If we do the conversion before the warning, then the user gets the "wxFormBuilder information" display with the non-wrapping, non-scrollable single line. The details button gives a bit more information, but it does not indicate that any kind of conversion is going on (in my test case it just says it is removing properties). I think we need to give the user some idea of what's going on before we display any kind of warning/error message that can result from the conversion.

What I have done is change the wording a bit to draw more attention to the fact that saving the file will change the project file.

BTW, I assume that you would prefer that I continue using a force push for any PR where I'm the only one making a change so that the final PR only shows the final change and not all the feedback changes?

sodevel:
It is not necessary that a PR consists of only one commit at the end, if there are multiple atomic operations that are worth to be kept for a meaningful history i prefer to preserve these. But cleanup commits or feedback changes or dramatic turnarounds in the implementation only clutter up the history and should be removed. While you are working on a PR it is not necessary to force push everything, i consider a PR private so you can do whatever you feel is reasonable. When i am about to merge it i might ask you to do a final cleanup.

src/rad/appdata.cpp
```diff
@@ -1380,15 +1373,14 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
 	return true;
 }
 
-bool ApplicationData::ConvertProject( const wxString& path, int fileMajor, int fileMinor )
+bool ApplicationData::ConvertProject(ticpp::Document& doc, const wxString& path, int fileMajor, int fileMinor)
 {
 	try
 	{
 		ticpp::Document doc;
 		XMLUtils::LoadXMLFile( doc, false, path );
```
sodevel: This can not and does not work (i actually tested this). The method parameter doc gets shadowed by the local one, your conversion gets lost at the end of the method.

Removing the shadowing is not enough, the calling code stores a reference to the root node of the document (Line 1275), but this method actually erases the document and rebuilds it so the reference of the outer code will break. This needs also to be addressed.

src/rad/appdata.cpp
```diff
@@ -1323,24 +1324,15 @@ bool ApplicationData::LoadProject( const wxString &file, bool justGenerate )
...
-				::wxCopyFile( file, file + wxT( ".bak" ) );
 
-				if ( !ConvertProject( file, fbpVerMajor, fbpVerMinor ) )
-				{
-					wxLogError( wxT( "Unable to convert project" ) );
-					return false;
-				}
+			wxMessageBox(
+			    _("This project is using an older file format. It will be converted in memory while loading it.\n\n"
+			      "Caution! If you save the project, it will be overwritten with the newer file format."),
```
sodevel: Better than before, but somehow i am still not happy about it. How about something like this:

This project file is using an older file format, it will be updated during loading.
WARNING: Saving the project will update the format of the project file on disk!

Randalphwa:
Unfortunately, I didn't get a notice from github when you added the last comment, so I didn't realize a change was still requested. I made the change, but now there is a merge conflict with the last PR. I merged this branch with the current origin\master which should include your recent commits as well as the last PR that touched rad/appdata.cpp.

sodevel:
Thanks for updating the message but you forgot to fix the variable shadowing so the project conversion itself is still broken (to be precise it is simply lost at the end of the method call).

Please don't do these kind of merges to fix merge conflicts. It creates a bad history and is quite hard to read, i actually had to merge your branch locally to actually see what it does. And then i saw something went wrong during your merges, you revert some of the line ending normalizations and file permission fixes by accident. If your feature branch does not merge please rebase it on current master.

Randalphwa:
I backed out the merge. I reset the root node pointer on successful conversion, and removed the bogus local doc class. Retested with an old project file to confirm that it writes the updated project when saved.

PR #574, opened 2019-12-13
Change handling of non-UTF8 XML files

Randalphwa:
Currently, when **wxFB** is asked to import an XRC file that uses an encoding other then `utf-8` it will ask the user for permission to backup the file and then save the original version with the encoding changed. However, it doesn't actually convert the strings in the file, it simply changes the encoding tag to `utf-8` and reloads the file. That means that not only will **wxFB** still not be able to correctly display non-ANSI characters, but by saving the file with the invalid encoding tag, the program that created it will also no longer be able to display strings correctly.

This PR changes the message displayed to the user to indicate strings may not be displayed correctly, and simply asks if the user wishes to continue or not. It removes the call to changing the encoding tag, and no longer saves and reloads the file.

Note that this is in the LoadXMLFileImp() function -- so it will affect any XML file loaded by **wxFB**. That should be fine -- if the user changed the encoding of a project file using an external XML editor, what they really need to do is fix the problem using that editor. **wxFB** does not have the ability to correctly convert the entire file to utf-8. It should not be saving a broken XML file that can no longer be fixed by any XML editor.

### DIFF
--- a/src/utils/stringutils.cpp
+++ b/src/utils/stringutils.cpp
@@ -179,22 +179,15 @@ namespace XMLUtils
 		else
 		{
 			// Ask user to all wxFB to convert the file to UTF-8 and add the XML declaration
-			wxString msg = wxString::Format( _("This xml file has specified encoding %s. wxFormBuilder only works with UTF-8.\n"),
-							wxFontMapper::GetEncodingDescription( encoding ).c_str() );
-			msg 		+= _("Would you like wxFormBuilder to backup the file and convert it to UTF-8\?\n\n");
-			msg			+= _("Path: ");
-			msg			+= path;
+			wxString msg = wxString::Format(_("This xml file has specified encoding: %s.\n\n"),
+			      wxFontMapper::GetEncodingDescription(encoding).c_str());
+			msg += _("wxFormBuilder may not correctly display strings unless the encoding is UTF-8.\n\n");
+			msg += _("Do you wish to continue?\n");
 			if ( wxNO == wxMessageBox( msg, _("Not UTF-8"), wxICON_QUESTION | wxYES_NO | wxYES_DEFAULT, wxTheApp->GetTopWindow() ) )
 			{
 				// User declined, give up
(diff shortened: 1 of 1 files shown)


sodevel:
I cannot confirm your observations. In a quick test a latin-1 encoded file got correctly converted to utf-8.

Looking at the source i can see how the file actually gets converted, in stringutils.cpp:329 the whole file gets read through a charset converter of the specified encoding and in stringutils.cpp:391 it gets written again in utf-8.

There are still two issues with the conversion code though, first one are these strange multiple recursive calls that read the file, this doesn't look quite right. The second issue is that the whole file gets read as the detected or specified encoding which is not correct because the xml preamble always is in ASCII. If the chosen encoding differs in the ASCII range from ASCII the preamble gets destroyed and the conversion code will fail.

Additionally your proposed workaround might lead to silent data loss, i remember wxWidgets string conversion functions have the habit to simply return an empty string if the conversion fails (well, technically empty string matches the description "not correctly display" :smile:).

Randalphwa:
Argh, I was focused on the fact that the encoding tag was being changed directly in the string instead of via the XML class, and completely missed the conversion on input. That still leaves the issue of changing the file on disk, which from the user's perspective is particularly odd since wxFB doesn't export the .xrc, it's just importing it into a project file. As you've probably noticed, I really hate having a repository changed just because I wanted to look at what someone else was doing. Another option would be to write and reread it from a temporary file -- that eliminates the need for a backup file, and doesn't risk changing a repository.

When I was experimenting with different approaches, I ran into that recursion bit when I discovered how easy it was to create an infinite recursion loop. I'll look into it further...

I'm going to come back to this in a bit -- the xrc conversion code also uses exceptions as the normal method of program control rather then checking the return value from tinyxml. That's generating hundreds of exceptions in pretty much any xrc file I've been looking at. I prefer breaking on any exception when running under a debugger, and that's no possible with the xrc files I'm looking at. And then there's the long list of memory leaks when exiting after importing one of these files...

PR #576, opened 2019-12-16
Add additional header symlinks for win32 build

fotocyte:
win32 builds complain about missing headers for bfd_stdint.h and
diagnostics.h


sodevel:
I remember that in my MSYS2 environment these extra links were required but for some reason on the CI using the same environment not. Since the switch to GitHub Actions PR's currently don't trigger the CI so i don't if this change does break the CI, maybe @jhasse can take a look at that first.

jhasse:
Hm ... why is CI not triggered for PRs?

The links weren't required on CI since it used an older MSYS2 snapshot.

sodevel:
Because if i understand the configuration file correct it is only enabled for push and not PR and i don't know if this is by intent or accident.

jhasse:
accident ;)

PR #580, opened 2020-01-02
Add cpp_preamble to inherited C++ code

Randalphwa:
It's unlikely that a user will want a precompiled header added to the generated C++ source file, and not want the same precompiled header added to the inherited C++ source file. This PR simply adds the same user-defined preamble to the inherited file(s) that is already added to the generated file(s).

### DIFF
--- a/output/xml/default.cppcode
+++ b/output/xml/default.cppcode
@@ -305,7 +305,13 @@
 			// Handlers for $basename events.
 		</template>
 		<template name="header_include">@#include &quot;$gen_file.h&quot;</template>
-		<template name="source_include">@#include &quot;$file.h&quot;</template>
+		<template name="source_include">
+			#ifnotnull $precompiled_header
+			@{
+				$precompiled_header #nl
+			@}
+			@#include &quot;$file.h&quot;
+		</template>
 		<template name="class_decl">
 			/** Implementing $basename */ #nl
 			class $name : public $basename
--- a/output/xml/default.xml
+++ b/output/xml/default.xml
@@ -227,6 +227,9 @@
     <property name="file" type="text"/>
     <property name="gen_file" type="text"/>
     <property name="type" type="text"/>
+    <category name="C++ Properties">
+      <property name="precompiled_header" type="text"/>
+    </category>
   </objectinfo>
 
   <objectinfo class="AUI" type="interface">
--- a/src/rad/appdata.cpp
+++ b/src/rad/appdata.cpp
@@ -2391,8 +2391,9 @@ void ApplicationData::GenerateInheritedClass( PObjectBase form, wxString classNa
 		PProperty fileProp = obj->GetProperty( wxT( "file" ) );
 		PProperty genfileProp = obj->GetProperty( wxT( "gen_file" ) );
 		PProperty typeProp = obj->GetProperty( wxT( "type" ) );
+		PProperty pchProp = obj->GetProperty(wxT("precompiled_header"));
 
-		if ( !( baseNameProp && nameProp && fileProp && typeProp && genfileProp ) )
+		if (!(baseNameProp && nameProp && fileProp && typeProp && genfileProp && pchProp))
 		{
 			wxLogWarning( wxT("Missing Property") );
 			return;
@@ -2425,6 +2426,12 @@ void ApplicationData::GenerateInheritedClass( PObjectBase form, wxString classNa
 		fileProp->SetValue( inherFile.GetName() );
 		genfileProp->SetValue( genFile.GetFullPath() );
 		typeProp->SetValue( form->GetClassName() );
+
+        auto property = project->GetProperty(wxT("precompiled_header"));
+        if (property)
+        {
+            pchProp->SetValue(property->GetValue());
+        }
 
 		// Determine if Microsoft BOM should be used
 		bool useMicrosoftBOM = false;
--- a/src/rad/appdata.cpp
+++ b/src/rad/appdata.cpp
@@ -2427,10 +2427,10 @@ void ApplicationData::GenerateInheritedClass( PObjectBase form, wxString classNa
 		genfileProp->SetValue( genFile.GetFullPath() );
 		typeProp->SetValue( form->GetClassName() );
 
-        auto property = project->GetProperty(wxT("precompiled_header"));
-        if (property)
+        auto pchValue = project->GetProperty(wxT("precompiled_header"));
+        if (pchValue)
         {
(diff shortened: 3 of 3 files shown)


sodevel:
I am not sure if a user wants the same precompiled header in the generated code and the generated implementation or if other users misuse this field to inject additional headers into the generated code that is not needed by the implementation like me 😄 . On the other hand i'm not really using the generated implementation because it just doesn't fit my needs and code style so i don't mind that addition.

But i do mind that you use the `cpp_preamble` template for this. Looking at the file `output/xml/default.cppcode` it looks like these templates are specifically designed to be used for the generated code and appear in pairs. Right now this template does not contain more then the inclusion of the precompiled header, but this might change in the future.

Further down in that file there is another "template-container" `UserClasses` that contains templates which are used by the code generator to generate the implementation. I think that addition would fit better there. Maybe it's not even necessary to introduce another template but just add a code fragment to the `source_include` template.

Randalphwa:
And here I thought this was going to be a really easy PR... :-)

So if I'm understanding the code correctly, prior to project file version 1.6, the precompiled_header property simply specified the name of the file. In version 1.6, it was decided that a preamble would be more appropriate allowing the use of a compiler flag (-DWX_PRECOMP) to determine which header file to include. Rather than adding a new property that could hold any preamble, the precompiled_header property got overwritten. As you noted, that's a problem because while it's likely that you would want the same precompiled header file in both files, that's not necessarily the case for a preamble which could effectively be anything.

I did try adding "$precompiled_header" to the "source_include" section of the "UserClasses" template, but that resulted in an error because "$precompiled_header" couldn't be found. I was able to create a "precompiled_header" section in the "Project" template class, but that just make that template confusing since then "precompiled_header" and "cpp_preamble" are identical.

Bottom line, though, is that without two project properties, I don't see any solution that isn't going to adversely affect someone. My inclination is to leave the original code as is. If at some point that cpp_preamble gets expanded functionality, and there is a project version change to match, then separate properties could be used and the issue revisisted. If you agree, then I think the PR should simply be closed.

sodevel:
> Bottom line, though, is that without two project properties, I don't see any solution that isn't going to adversely affect someone.

But i do, but this still conflicts with your really easy PR intention. The reason why using $precompiled_header inside the template didn't work is because the template parser gets passed a UserClasses object which doesn't contain this property, it is part of the Project object. But the template parser does require that UserClasses object to use its properties like $file. I see 4 possible solutions for this problem:

1. Extend the template parser to accept two objects for property extraction: Lots of work, doesn't seem to be needed anywhere else, what to do if both objects contain properties with the same name?
2. Add extra template to UserClasses and pass that with the Project object to the template processor: Quite confusing that this template gets a different object to resolve its properties resulting in no UserClasses properties to work.
3. Add precompiled_header property to UserClasses, maybe inside a group like Project does itself (see `output/xml/default.xml`) and fill that property manually like all UserClasses properties are filled (see `ApplicationData::GenerateInheritedClass()`): Allows the usage of all UserClasses properties plus that special one Project property.
4. Extract the precompiled_header property directly from the Project object and insert it in literal: This does work because it is a pure text property but totally circumvents the value-to-code processing of TemplateParser, this smells like a huge hack.

Solution 3 seems to be the most appropriate to me.

Randalphwa:
Okay, I've added **precompiled_header** as a property in default.xml, and a template in default.cppcode, filled it in in appdata, and replaced the cppcg.cpp code. I'm still chuckling about thinking this would be a simple 3-line code fix...

src/rad/appdata.cpp
```diff
@@ -2391,8 +2392,9 @@ void ApplicationData::GenerateInheritedClass( PObjectBase form, wxString classNa
 		PProperty fileProp = obj->GetProperty( wxT( "file" ) );
 		PProperty genfileProp = obj->GetProperty( wxT( "gen_file" ) );
 		PProperty typeProp = obj->GetProperty( wxT( "type" ) );
+		PProperty typePCH = obj->GetProperty(wxT("precompiled_header"));
```
sodevel: Following the present naming style `pchProp` makes more sense and looks less confusing

src/rad/appdata.cpp
```diff
@@ -2367,6 +2367,7 @@ void ApplicationData::GenerateCode( bool panelOnly, bool noDelayed )
 	NotifyCodeGeneration( panelOnly, !noDelayed );
 }
 
+// Properties in ../output/xml/default.xml and code templates in ../output/xml/default.cppcode
```
sodevel: This comment is a bit misleading. This is generic code and applies to all languages, not only c++. Why is that path relative to `src`?

output/xml/default.cppcode
```diff
@@ -304,6 +304,12 @@
 		<template name="event_handler_comment">
 			// Handlers for $basename events.
 		</template>
+		<template name="precompiled_header">
+			#ifnotnull $precompiled_header
+			@{
+				$precompiled_header
+			@}
+		</template>
```
sodevel: Regarding the position where that pch gets inserted i don't see a reason to use an extra template for that, can't it be just prepended to the `source_include` template?

src/codegen/cppcg.cpp
```diff
@@ -457,6 +457,13 @@ void CppCodeGenerator::GenerateInheritedClass( PObjectBase userClasses, PObjectB
 	m_header->Indent();
 
 	// Start source file
+	code = GetCode(userClasses, wxT("precompiled_header"));
+    if (!code.empty())
+    {
+        m_source->WriteLn(code);
+        m_source->WriteLn(wxEmptyString);
+    }
+
```
sodevel: This could be removed if only the `source_include` template would be used

output/xml/default.xml
```diff
@@ -227,6 +227,7 @@
     <property name="file" type="text"/>
     <property name="gen_file" type="text"/>
     <property name="type" type="text"/>
+    <property name="precompiled_header" type="text"/>
```
sodevel: Can this property be wrapped inside a category `C++ Properties` like it's done in the Project template? These properties are not displayed anywhere currently but it would make the xml more clear that this is a C++ only property.

src/rad/appdata.cpp
Randalphwa: My bad, I had that in there just to make it easier to switch between the three files in the editor. No reason for it to be in the PR.

Randalphwa:
I made the requested changes, and also changed to 'auto property =' for retrieving the value instead of `PProperty pPCH =' in appdata.cpp.

PR #581, opened 2020-01-11
add Cmake build files

### LINKED ISSUE #264: CMake Build System
I would like to know if anybody is currently working on CMake build for wxFormBuilder, which would probably improve Windows and XCode build experience.

There seems to be an existing fork from before the move to github at:
https://github.com/marekr/wxFormBuilder

There has been talk about it at the mailing list:
https://sourceforge.net/p/wxformbuilder/mailman/wxformbuilder-devel/

And it has been mentioned in #246

abma:
fixes #264 

its far away from beeing perfect but IMHO its a good start as only cmake files are added.

works for me on debian 10 buster.

in the original cmake_build branch some non-cmake files were changed: i've reverted these changes.


sodevel:
This branch is stale for quite some time now. In the long shot my plan is to move to CMake, but right now i know basically nothing about it. I know there is something called "Modern CMake" but i can't judge how much this branch already adheres that.

There are currently 3 build systems in place, each with their own flaws and limitations, i don't know if it is a good idea to add a 4th. Checking the files i spotted some hardcoded paths for at least the MinGW route so this is the first that won't work.

I would prefer some input from someone else who knows more about CMake than me. Maybe someone can try to use CMake on the CI to check how well it works there.

abma:
> There are currently 3 build systems in place, each with their own flaws and limitations, i don't know if it is a good idea to add a 4th. Checking the files i spotted some hardcoded paths for at least the MinGW route so this is the first that won't work.

adding a 4th wouldn't be good: should all others being dropped in this pull request?
> Checking the files i spotted some hardcoded paths for at least the MinGW route so this is the first that won't work.

where did you spot some hardcoded paths?

> I would prefer some input from someone else who knows more about CMake than me. Maybe someone can try to use CMake on the CI to check how well it works there.

i can try to add ci support for cmake: but then again some common files have to be touched which may lead to merge conflicts when this pull request isn't merged in a timely manner.

sodevel:
> adding a 4th wouldn't be good: should all others being dropped in this pull request?
> 
This PR should not touch any of the other build systems, this is just out of its scope. The reason for the multiple build systems is that a single one can not build all the different variants. The matrix looks like this:
* Windows
  * MinGW (MSYS2)
  * MSVC
* Linux
  * GCC flatpak
  * GCC native
* MacOS
  * GCC AppPackage

Another problem is that on Windows still wxWidgets 3.0.4 is used, mostly because MSYS2 doesn't offer a newer package yet and MSVC isn't build on CI. wxFB has quite some issues with that version, the Linux and MacOS variants use newer versions through homebrew/git, it would be great if that could be done on Windows as well.

> where did you spot some hardcoded paths?
> 
`src/CMakeLists.txt:128`
 
> i can try to add ci support for cmake: but then again some common files have to be touched which may lead to merge conflicts when this pull request isn't merged in a timely manner.
>
There is currently no ongoing work on the CI configuration beside my tests to replace Premake 4 with Premake 5. I would also prefer to replace Travis with GitHub Actions but i won't work on that in the near feature. Therefor the risk of conflicts is low, especially if you would use GitHub Actions. But even if not, these should not be hard to resolve.

abma:
cmake should support all buildsystems you listed.

flatpak supports cmake, too. but when implementing these, the existing files have to be touched/changed. :-|

abma:
wrt the hardcoded path: can the msys path be detected by an env var or registry setting?

i.e. https://stackoverflow.com/questions/1762201/reading-registry-values-with-cmake

sodevel:
> flatpak supports cmake, too. but when implementing these, the existing files have to be touched/changed. :-|

Which existing files? Can't you just add additional files with sensible names that can co-exists with the present ones?

> wrt the hardcoded path: can the msys path be detected by an env var or registry setting?

I don't know, but does this still apply to Modern CMake? I did a bit of research and as far as i can tell the current state of the CMake files is far away from beeing modern. Especially the use of build-requirements and usage-requirements from 3.12 onwards could be of great benefit for the plugins and sdk.

Wouldn't it be a better a approach to just start a new branch from scratch instead of trying to update the present one?

Randalphwa:
Just as an FYI, I made a local copy of the PR and loaded up src\cmakelists.txt in the latest Visual Studio 2019. It generated the following warnings and error:

 [CMake] CMake Warning (dev) in C:\rwCode\ttForm\src\CMakeLists.txt:
1> [CMake]   No project() command is present.  The top-level CMakeLists.txt file must
1> [CMake]   contain a literal, direct call to the project() command.  Add a line of
1> [CMake]   code such as
1> [CMake] 
1> [CMake]     project(ProjectName)
1> [CMake] 
1> [CMake]   near the top of the file, but after cmake_minimum_required().
1> [CMake] 
1> [CMake]   CMake is pretending there is a "project(Project)" command on the first
1> [CMake]   line.
1> [CMake] This warning is for project developers.  Use -Wno-dev to suppress it.
1> [CMake] 
1> [CMake] CMake Error at C:\rwCode\ttForm\src\CMakeLists.txt:179 (install):
1> [CMake]   install DIRECTORY given no DESTINATION!
1> [CMake] 
1> [CMake] 
1> [CMake] CMake Warning (dev) in C:\rwCode\ttForm\src\CMakeLists.txt:
1> [CMake]   No cmake_minimum_required command is present.  A line of code such as
1> [CMake] 
1> [CMake]     cmake_minimum_required(VERSION 3.15)
1> [CMake] 
1> [CMake]   should be added at the top of the file.  The version specified may be lower
1> [CMake]   if you wish to support older CMake versions for this project.  For more
1> [CMake]   information run "cmake --help-policy CMP0000".
1> [CMake] This warning is for project developers.  Use -Wno-dev to suppress it.
1> [CMake] 
1> [CMake] -- Configuring incomplete, errors occurred!

abma:
@Randalphwa: you have to run cmake in the project root dir: 

https://github.com/wxFormBuilder/wxFormBuilder/blob/d49a6041f9f5525737917fa79c63852db788085f/CMakeLists.txt

Randalphwa:
Hmm, okay. Seems odd that you can't be in the directory of the sub-project you want to work on if you want to use cmake-functionality. That must be why I so rarely use it... :-)

I tried it in the root, and that made the underlying cmake happy. The build itself fails, though that appears to be configuration issues on my end. Will track down why my private builds work and this one doesn't and let you know if there's an underlying issue with Visual Studio and this cmakelist.txt.

.gitignore
```diff
@@ -24,3 +24,13 @@ install/windows/*.exe
 /_install/
 /compile_commands.json
 /.ccls-cache/
+
+# CMake
+CMakeFiles
```
jhasse: This encourages in-source buliding. I would recommend people to use `_build/` as a build dir (or just `build` when Premake is removed).

.travis.yml
```diff
@@ -1,36 +1,17 @@
...
+    - dist: bionic
+      addons:
+        apt:
+          packages:
+            - libwxgtk3.0-dev
+            - libwxgtk-media3.0-dev
+            - g++-8
+            - cmake
+      install: export CXX="g++-8" CC="gcc-8"
+      before_script: cmake .
```
jhasse: `cmake -B_build -H.`

.travis.yml
```diff
@@ -1,36 +1,17 @@
...
+      script: make
```
jhasse: `cmake --build _build`

(that will work with any generator, e.g. Xcode on macOS)

CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+project(wxFormBuilder)
+cmake_minimum_required(VERSION 3.0)
+set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMakeModules)
+set(CMAKE_CXX_STANDARD 17)
```
jhasse: Use `target_compile_features(myTarget PUBLIC cxx_std_17)` instead, see https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html

CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+set(CPACK_WIX_UPGRADE_GUID B5AAB525-34D5-4B12-B6F2-136C8ED890C9)
+
+if(APPLE)
+    set(CPACK_SYSTEM_NAME macOS)
+endif()
+
+include(CPack)
+
+if(MSVC)
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
jhasse: Has been superseded by `add_compile_definitions`, see https://cmake.org/cmake/help/v3.16/command/add_definitions.html

src/CMakeLists.txt
```diff
@@ -0,0 +1,183 @@
...
+            ${BFD_LIB}
+            ${IBERTY_LIB}
+            ${Intl_LIBRARIES}
+            ${ZLIB_LIBRARIES}
+            psapi
+            imagehlp)
+    endif()
+endif()
+
+if(CMAKE_GENERATOR MATCHES "^MSYS")
```
jhasse: Hm ... what if someone uses Ninja on MSYS2?

abma: i don't use msys, so no clue.

its a regex, so it should match MSYS and MSYS2

CMakeLists.txt
abma: which cmake version is targeted? i also don't use MSVC, so its stupid when i change it, as i can't test it

.gitignore
abma: thats a personal preference IMHO, i prefer in source builds: this line really doesn't hurt.

abma:
please merge or i'm out as i only can test on debian.

abma:
@jhasse 
i have no clue why the travis build fails. locally the commands work in a clean checkout. as you already gave most hints / change requests IMHO its better when you finish this as basicly all changes you requested where one-liners: so its easier when you change it directly instead of writing it to me and i just paste it...

abma:
see #631

PR #583, opened 2020-01-25
Inherited code generation

nathanjshaffer:
Clean up inherited code generation.  Doesn't mangle indenting and retains source file organization#469

### DIFF
--- a/src/codegen/codeparser.cpp
+++ b/src/codegen/codeparser.cpp
@@ -44,7 +44,8 @@ wxString Function::GetFunction() {
 // CodeParser
 //---------------------------------------------------
 
-void CCodeParser::ParseCFiles(wxString className) {
+void CCodeParser::ParseFiles(wxString className) {
+
 	m_className = className;
 	wxTextFile headerFile(m_hFile);
 	wxTextFile sourceFile(m_cFile);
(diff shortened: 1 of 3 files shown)


nathanjshaffer:
> This is a tough one, actually this is the first time that im looking at these code parts, there are a lot of issues, not only in this PR but the original code as well. Biggest issue is the memory leak that the original code sort of works around with `GetFunctionContents()`, but because this PR removes all but one call of this method the leak gets hit every time. Second biggest problem is the unnecessary copying of many parameters instead passing them as const-ref.
> 
> So far i can see this PR adds an array to store the order of the functions that are stored in a hashmap. That hashmap is the cause of the memory leak because it stores its contents as heap objects which don't get deleted anymore (except for the constructor function of the generated class) because of the changed logic. As far as i can see in the new code there is only one function looked up by its identifier, that constructor function. All other functions are processed in order. I think for this workflow that hashmap isn't required anymore, all functions can be stored in the array, searching through that array one time shouldn't hurt much. This would solve the memory leak and also simplify the code.
> 
> I'm trying to mark most of the other issues inline.

Thanks for the feedback. I have made the requested changes and tighned everything up.  There is now just a custom wxArray object that stores all of the user code and memory is cleaned up when the array goes out of scope.  I have also changed all the variable/function naming as suggested.

sodevel:
Just a quick update from my side, i saw the update but i am currently too busy for a detailed review, i will look at it as soon as possible.

sodevel:
Sorry for the long delay but i was and still am very busy with other tasks so i can't devote much time to wxFormBuilder currently.

Thanks for working on this, it is an improvement but the main problem still exists: the memory leak. This is because you are storing raw pointers in a container. Clearing the container deletes the pointers but not what they point to. In general don't store raw pointers inside containers. Also don't use wxArray, it is deprecated and should not be used anymore. And don't try to use its object variant that tries to support raw pointers.

If i understand the code right the only reason to use pointers is to model the attribute "not present" for an item. It is a bad design choice to use pointers for that, there are other solutions available but im kind of short of time to explain them. A quickfix to solve the memory leak would be to use std::shared_ptr for the array contents (and to replace the array with std::vector). A better approach would be to use std::optional but even better, but i think it would be even better to change the interface to prevent creating unneccessary copies.

src/codegen/codeparser.cpp
```diff
@@ -81,119 +86,157 @@ void CCodeParser::ParseCFiles(wxString className) {
...
+		{
+			// find the end of the user include
+			wxString Str = m_className + wxT("::");
+			userIncludeEnd = code.find(Str, userIncludeStart);
+			if (userIncludeEnd != wxNOT_FOUND) 
+			{
+				userIncludeEnd = code.rfind(wxT('\n'), userIncludeEnd);
+				if (userIncludeEnd != wxNOT_FOUND) 
+				{
+					userIncludeStart++;
```
sodevel, line 144: Always use pre-increment unless you really need the post-increment

src/codegen/codeparser.cpp
```diff
@@ -81,119 +86,157 @@ void CCodeParser::ParseCFiles(wxString className) {
...
-	m_userMemebers = wxT("");
-	int userMembersStart = code.Find(wxT("//// end generated class members"));
+void CppCodeParser::ParseHeaderFileUserMembers(wxString &code) {
+	m_userHeaderFileMembers = wxT("");
+	int userMembersStart = code.Find(wxT("//// end generated class members. Place User Code Below"));
 	if (userMembersStart != wxNOT_FOUND) {
 		userMembersStart = code.find('\n', userMembersStart);
 		if (userMembersStart == wxNOT_FOUND) {
-			m_userMemebers = wxT("");
+			m_userHeaderFileMembers = wxT("");
```
sodevel, line 168: Either assign wxEmptyString or call clear()

src/codegen/codeparser.cpp
```diff
@@ -81,119 +86,157 @@ void CCodeParser::ParseCFiles(wxString className) {
...
 		}
 	}
 }
 
-void CCodeParser::ParseCUserMembers(wxString code) {
-	m_userMemebers = wxT("");
-	int userMembersStart = code.Find(wxT("//// end generated class members"));
+void CppCodeParser::ParseHeaderFileUserMembers(wxString &code) {
+	m_userHeaderFileMembers = wxT("");
+	int userMembersStart = code.Find(wxT("//// end generated class members. Place User Code Below"));
```
sodevel, line 164: Only use the wxWidgets String-Methods if there is no std::string variant available. The wxWidets methods start with uppercase letters while the std::string interface ones start with lowercase letters.

src/codegen/codeparser.cpp
```diff
@@ -81,119 +86,157 @@ void CCodeParser::ParseCFiles(wxString className) {
...
+	while (functionStart != wxNOT_FOUND) {
 		// find the begining of the function name
 		Str = m_className + wxT("::");
 		functionStart = code.find(Str, previousFunctionEnd);
 		if (functionStart == wxNOT_FOUND) {
 			// Get the last bit of remaining code after the last function in the file
-			m_trailingCode = code.Mid(previousFunctionEnd);
-			m_trailingCode.RemoveLast();
+			if(code.GetChar(previousFunctionEnd) == '\n')
+				{previousFunctionEnd++;}
```
sodevel, line 196: This brace placement is bad, either use the Java style with brace on same line or place it on the next line, even for one-liners please.

src/codegen/codeparser.h
```diff
@@ -58,98 +62,132 @@ class Function {
...
-		m_cFile = sourceFileName;
-	}
+	/** extracts each function in source File**/
+	void ParseSourceFileFunctions(wxString &code);
+
+	wxString ParseBrackets(wxString &code, int &functionStart);
+
+	// Accessor Functions
+	////////////////////////
+	wxString GetUserHeaderFilePreProc() {return m_userIncludeFilePreProc;}
```
sodevel, line 149: For simple accessors one-liners are ok but please add whitepace inside the braces

src/codegen/codeparser.h
```diff
@@ -58,98 +62,132 @@ class Function {
 	wxString m_documentation;
 };
 
-/** map class mapping Function* to function name */
-using FunctionMap = std::unordered_map<std::string, Function*>;
-#define funcIterator FunctionMap::iterator
+
+WX_DEFINE_ARRAY(CppFunction*, ArrayCppFunctionBase);
```
sodevel, line 66: Don't use deprecated wxArray, use std::vector

PR #585, opened 2020-01-28
Removed deprecated function AddLabelTool

osnipezzini:


sodevel:
The commit message is kind of misleading because it does not remove something but rather replace something. But I am going to merge this as-is and use a more fitting message for the merge commit. For the next time please use https://chris.beams.io/posts/git-commit/ as a guidance how to write commit messages.

PR #592, opened 2020-02-26
Fix build errors for windows 64-bit

tomay3000:


src/dbg_stack_trace/stack.cpp
```diff
@@ -293,7 +300,12 @@ namespace
 
                 if (func.empty())
                 {
-                    DWORD displacement = 0; // dummy variable
+                    #if defined(__amd64__) || defined(__x86_64__) // 64-bit
+                        DWORD64 displacement = 0; // dummy variable
+                    #elif defined(__i386__) || defined(_X86_) // 32-bit
+                        DWORD displacement = 0; // dummy variable
+                    #endif
```
sodevel: In the `#else` case right below (i can't mark it so i mark this block instead) there is that very same `displacement` variable, shouldn't its type changed depending on architecture like in this block?

tomay3000: You are right, I think it should, I haven't notice it, because I compiled my forked repo using MinGW-w64 that is the `#if defined(__MINGW32__)`. If using `MSVC`, then the `#else` `DWORD displacement = 0; // dummy variable` should be patched too using `MSVC` specific pre-processor, because I don't think we are targeting `ARM` and not known compiler for the moment.

tomay3000: Alongside with `MinGW-w64`, I have only `Visual Studio 2010` installed which have no `C++11` support to compile, I need a `Visual Studio 2015+` tester, or I am gonna apply the missing fixes anyway!

sodevel: There is currently no easy way to make a 64 bit build with MSVC, i'm trying to work on that, then i can build with MSVC 2019

sodevel:
Sorry for the long delay, didn't have time to take care about this earlier.

I have a test setup for 64 Bit builds with MSVC now, your changes compile but produce two warnings that should lead to a crash when this code gets actually executed.

````
src\dbg_stack_trace\stack.cpp(293,31): warning C4244: 'initializing': conversion from 'DWORD64' to 'DWORD', possible loss of data
src\dbg_stack_trace\stack.cpp(295,15): warning C4312: 'reinterpret_cast': conversion from 'DWORD' to 'HINSTANCE' of greater size
````

The second one should be void if the first one gets fixed, seems like the module address turns into a DWORD64 on 64 bit builds.

tomay3000:
Mmm, I have missed that one too, I will add another commit to this soon.
Seems like `AppVeyor` has becoming a must now.

sodevel:
This is looking good now, although i was not able to actually do something that does produce a stack trace, the code looks good.

Can you please merge these commits into one and write a meaningful commit message and force-push it so that i can just do a simple merge?

tomay3000:
> Can you please merge these commits into one

Yeah, sure.

> and write a meaningful commit message

Like what!? I thought I did my best for this one, English is my third official language not even the second.

tomay3000:
I have merged all the commits in a single one (with a history rewrite and a forced-push).

PR #596, opened 2020-03-16
Fix #374: In case of a system dark theme SetPropertyTextColour to black

### LINKED ISSUE #374: UI problem with system Dark Theme
Look at the screenshot please:

http://oi68.tinypic.com/2w7jui1.jpg

It's because I'm using DARK THEME for my Windows OS(7) and texts colors in Properties panel are Hardcoded... So wxFormBuilder is not able to use native/system color

Both text and background are white

michallukowski:
When the system has a dark theme, the default font is light and cannot be seen on light backgrounds

### DIFF
--- a/src/rad/inspector/objinspect.cpp
+++ b/src/rad/inspector/objinspect.cpp
@@ -528,11 +528,20 @@ void ObjectInspector::AddItems( const wxString& name, PObjectBase obj,
 			{
 				// Most common classes will be showed with a slightly different colour.
 				if (name == wxT("wxWindow"))
+				{
 					m_pg->SetPropertyBackgroundColour(id,wxColour(255,255,205)); // yellow
+					m_pg->SetPropertyTextColour(id, *wxBLACK);
+				}
 				else if (name == wxT("AUI"))
+				{
 					m_pg->SetPropertyBackgroundColour(id,wxColour(240,240,255)); // light blue
+					m_pg->SetPropertyTextColour(id, *wxBLACK);
+				}
 				else if (name == wxT("sizeritem") || name == wxT("gbsizeritem") || name == wxT("sizeritembase") )
+				{
 					m_pg->SetPropertyBackgroundColour(id,wxColour(220,255,255)); // cyan
+					m_pg->SetPropertyTextColour(id, *wxBLACK);
+				}
 			}
 
 			ExpandMap::iterator it = m_isExpanded.find( propName );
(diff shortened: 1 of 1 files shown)


sodevel:
Sorry for the late reply, didn't had time earlier.

I don't like these changes because they make the situation worse, instead of hardcoding one colour value you now hardcode two colour values, i guess it's only a matter of time until someone complains that his text colour doesn't match anymore. I would suggest a more dynamic approach to solve this issue.

Depending on the lightness of the font colour either

1. Choose a darker background colour
2. Adjust or invert the font colour to produce a better contrast to the background colour

This should produce better results across different themes.

michallukowski:
> I don't like these changes because they make the situation worse,

I can not agree. In the current situation, in some cases the font on yellow, light blue and cyan background is almost illegible. I attached a screenshot of what it looks like on my system. On the left is the current version of wxformbuilder, on the right is wxformbuilder with my changes.

tark_theme_solution

> Choose a darker background colour

I agree with it. Darker yellow, light blue and cyan backgrounds should look better depending on system themes. If I find some time, I will try to suggest a solution.

Regards

tomay3000:
Maybe you can use the recently added `double wxColour::GetLuminance() const` in `wxWidgets v3.1.3` [https://docs.wxwidgets.org/trunk/classwx_colour.html#ab26df3bfab77f5a3c54e9caff93c78f8](https://docs.wxwidgets.org/trunk/classwx_colour.html#ab26df3bfab77f5a3c54e9caff93c78f8) to decide whether the color is lighter or darker.

sodevel:
Keep in mind though that there is still only wxWidgets 3.0.4 available for MSYS2 and this is still the main platform to build for windows. The mentioned method however is not that complicated to be rewritten.

Looking at the screenshots i would vote for trying approach 1, use a darker set of background colours if the text color is too light, this should fit much better into a dark theme.

tomay3000:
> Keep in mind though that there is still only wxWidgets 3.0.4 available for MSYS2 and this is still the main platform to build for windows.

Yes, but sooner will be `v3.1.x` available, it is right there [https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-wxWidgets3.1](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-wxWidgets3.1) only not built yet, anyone can easily build and install it like so:
```sh
clone https://github.com/msys2/MINGW-packages.git
cd mingw-w64-wxWidgets3.1
MINGW_INSTALLS=mingw64 makepkg-mingw -sLf
pacman -U mingw-w64-wxWidgets3.1*.pkg.tar.zst
```
I personally use the official `wxWidgets v3.1.3` pre-built binaries and `MinGW-w64 v8.1.0` with only a few `Search and Replace` using only `Notepad++` or `sed` of the generated `.make` files, and I am using the compiled version of `wxFormBuilder x64` without problems so far.

> The mentioned method however is not that complicated to be rewritten.

Yes, indeed, it is defined here: [https://github.com/wxWidgets/wxWidgets/blob/master/src/common/colourcmn.cpp](https://github.com/wxWidgets/wxWidgets/blob/master/src/common/colourcmn.cpp) like this:
```cpp
double wxColourBase::GetLuminance() const
{
    return (0.299*Red() + 0.587*Green() + 0.114*Blue()) / 255.0;
}
```

> Looking at the screenshots i would vote for trying approach 1, use a darker set of background colours if the text color is too light, this should fit much better into a dark theme.

I think so too.

sodevel:
Because there was no further progress i implemented the dynamic approach myself and extended it to a few other elements that also did not adapt well in a dark theme. Depending on the actual theme sometimes this still is not ideal, e.g. the Mate BlackMATE uses black and white text, but with Adapta it is looking quite good.

PR #597, opened 2020-04-02
Symbolic link - removed lines where files not found

va6un:
These files /mingw32/include/binutils/**diagnostics.h**, /mingw32/include/binutils/**bfd_stdint.h** where not found.


sodevel:
I don't use MSYS2 myself and this is just a documentation update so it doesn't do anything to the CI that uses  MSYS2 so i don't what does this change do? Why is it necessary? And especially why does it work on the CI without that change? Here are certain issues that report compilation problems with MSYS2 and one issue was the missing link to diagnostics.h that this PR removes again.

I need more details why this PR is necessary and what specific environment is affected by that.

sodevel:
Current GitHub Runners use more recent MSYS2 distributions that require even more symlinks to be able to use binutils with default paths. Therefor im closing this PR, these changes won't work with current MSYS2 releases.

PR #604, opened 2020-05-07
Update wrong dataview property prefix

osnipezzini:
Add prefix to python code generations :

DATAVIEW_CELL_INERT,
DATAVIEW_COL_HIDDEN,
DATAVIEW_COL_RESIZABLE,
DATAVIEW_COL_SORTABLE,
DATAVIEW_CELL_ACTIVATABLE,
DATAVIEW_COL_REORDERABLE


src/codegen/pythoncg.cpp
```diff
@@ -1859,6 +1859,13 @@ void PythonTemplateParser::SetupModulePrefixes()
 	ADD_PREDEFINED_PREFIX( wxTL_3STATE, wx.dataview. );
 	ADD_PREDEFINED_PREFIX( wxTL_USER_3STATE, wx.dataview. );
 	ADD_PREDEFINED_PREFIX( wxTL_DEFAULT_STYLE, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxTL_DEFAULT_STYLE, wx.dataview. );
```
sodevel, line 1862: This line is a copy of the one directly above, remove it please

src/codegen/pythoncg.cpp
```diff
@@ -1859,6 +1859,13 @@ void PythonTemplateParser::SetupModulePrefixes()
...
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_CELL_INERT, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_COL_HIDDEN, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_COL_RESIZABLE, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_COL_SORTABLE, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_CELL_ACTIVATABLE, wx.dataview. );
+	ADD_PREDEFINED_PREFIX( wxDATAVIEW_COL_REORDERABLE, wx.dataview. );
```
sodevel, line 1868: These symbols of two enums are randomly listed, i think it would be better to sort them alphabetically, this would look cleaner.

osnipezzini:
Fixed commits

sodevel:
I would have kept these column-based values in the previous location behind the general dataview elements, but i can live with the current location as well. Anyway, can you squash all these commits into one and force-push that one so that i can merge this PR? I could squash them myself during merge but then the branch information gets lost.

osnipezzini:
Problem is , i know just basics of git commands , so i need to delete repo and fork again to send a clean pr

sodevel:
Oh no, please don't delete anything. Ok, i see the problem, you have used your master branch as base for this PR, this is a bad approach and will give you trouble once you need to synchronize your fork with this repo again.

In general use a separate branch for your changes and use that as base for the PR, if you would have done it like this, a cleanup would be a simple `git rebase -i master`. Anyway, this won't work in the current case, i am going to squash your commits during merge.

osnipezzini:
Thanks so much for understand , i'm really appreciate

PR #611, opened 2020-05-29
Gave information about what wxFormBuilder is about

ViniDalvino:


sodevel:
There are several issues with this PR why i can't accept it in the current state.

It contains links to images hosted at an external free location which is beyond the control of this project. This Readme file has a high visibility and that hosting is pretty unreliable and know to break sooner than later. If you want to reference static resources like these it is much better to include them in this repository and link to them, this is a much more reliable solution.

The additions contain pretty obvious orthographic errors and the language style is quite poor, this is unacceptable for the presentation and has to be improved.

sodevel:
Closing this because of inactivity. The updated README superseeds this as well.

Issue #524, opened 2019-06-10
Run Error on ubuntu 18.04


oxarbitrage:
Afer installation, on my Linux Mint 19.1(Ubuntu 18.04 based) there is a runtime error with the libs:

Could not set current working directory (error 2: No such file or directory)

This is similar to https://github.com/wxFormBuilder/wxFormBuilder/issues/285 but a bit different.

Libraries were built at:
_install/lib/x86_64-linux-gnu/wxformbuilder

While the program expect them to be just at:
_install/lib/wxformbuilder

st3255:
Same problem here, also not loading any widget content etc. Manually copying the files to the expected path (from /x86_64-linux-gnu/wxformbuilder to /wxformbuilder) solves this.

elig0n:
For anyone else interested, after `meson _build --prefix $PWD/_install`, from the repo's root run:
`cp -r _install/lib/x86_64-linux-gnu/wxformbuilder/ _install/lib/` and then run as usual i.e. `ninja -C _build install`

jmoraleda:
The instructions that worked for me to work around this issue were, from the `git` checkout directory:
* Configure and build the usual way:
`meson _build --prefix $PWD/_install`
`ninja -C _build install`
* Add a symbolic link to the location where the executable is expecting the libraries
`ln -s x86_64-linux-gnu/wxformbuilder/ _install/lib`
* Run the usual way
`./_install/bin/wxformbuilder`

sodevel:
The root cause is that newer versions of meson / different platforms it runs on output libraries in a architecture tagged subdirectory. I don't know why it does this, how it can be changed or if this is state-of-the-art now and wxFB should be changed.

jmoraleda:
FWIW I just read that meson's default `libdir` is platform dependent (https://mesonbuild.com/Builtin-options.html). I don't know what the best way to deal with the problem is.

jhasse:
I think wxFormBuilder should also check the platform-dependend directory, not sure what's the best way to get it in C++.

sodevel:
[...]

PR #613, opened 2020-05-30
Package names update in Ubuntu

guysoft:
According to Ubuntu 20.04, these are the packages that are needed now, otherwise you get:
```
E: Package 'libwxgtk3.0-dev' has no installation candidate
E: Unable to locate package libwxgtk-media3.0-dev
E: Couldn't find any package by glob 'libwxgtk-media3.0-dev
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ cd ../../../output/
 Pre-requisites for Ubuntu:
 
 ```sh
-sudo apt install libwxgtk3.0-dev libwxgtk-media3.0-dev meson
+sudo apt install libwxgtk3.0-gtk3-dev libwxgtk-media3.0-gtk3-dev meson
 ```
 
 Pre-requisites for Arch Linux:


damufo:
Hello:

I followed the installation @guysoft  instructions.
Once compiled I had to make a link from 
/home/user/dev/wxFormBuilder/_install/lib/wxformbuilder/
-->
/home/user/dev/wxFormBuilder/_install/lib/x86_64-linux-gnu/wxformbuilder/
https://imgur.com/8qIdQkS.png

Another detail, every time I boot the logging window appears Logging dialog
https://imgur.com/Ca04Wq4.png
How do I prevent this window from appearing?

Very thanks!

guysoft:
Hey @damufo we might need to add this line too. I had the same issue here and this is the workaround:
```
ln -s /home/guy/tmp/wxFormBuilder/_install/lib/x86_64-linux-gnu/wxformbuilder /home/guy/tmp/wxFormBuilder/_install/lib/wxformbuilder
```

So the full part is:
```
meson _build --prefix $PWD/_install
ninja -C _build install
ln -s /home/guy/tmp/wxFormBuilder/_install/lib/x86_64-linux-gnu/wxformbuilder /home/guy/tmp/wxFormBuilder/_install/lib/wxformbuilder
```

run:
```
./_install/bin/wxformbuilder
```

damufo:
Yes, that’s exactly what I did.
It worked for me.
What I don’t know is how to make the logging window not appear every time I start wxFormbuilder.
https://imgur.com/Ca04Wq4.png

guysoft:
@damufo I don't know regarding the debug window, but I guess that means you successfully using the instructions I provided :)
You might want to open a separate issue on that.

damufo:
OK

sodevel:
The documentation currently does not name a specific Ubuntu release it applies to and currently refers to package names that have been used in the past. Apparently these have changed in current releases, maybe in addition to the updated package names the minimum release which uses these names should be added as well.

About the not found libraries, this is a known problem, see #524. There is currently not a good solution for this because this is currently only caused by using Meson.

That logging dialog is only displayed by a debug build so for some reason you made a debug build instead of a release build. What were the exact command lines to generate the build files and executing the build?

guysoft:
I suggest you decide either to have packages according to the latest Ubuntu LTS (long term support). Or the latest. In both cases its 20.04, which uses the package in the PR. IMHO not updating would give new people installing a hard time for no good reason other than that the pages used to be called that way. 
Again, this what I would warmly suggest regarding this PR, not the library paths. It kinda went off topic really.

damufo:
meson _build --prefix $PWD/_install --buildtype=release

Work fine for me.
Very Thanks!!

igor-cali:
For those who are interested, I successfully created a deb package for Focal Fossa [20.04] ([personal sandbox ppa here](https://launchpad.net/~igor-cali/+archive/ubuntu/junk)).
I plan to review the package in July when the 20.04.1 release is available (I usually upgrade my machine when LTS upgrade are rolled out). After that, I could release the package in my stable ppa.

Issue #618, opened 2020-07-06
Create RichText XRC object for RichTextCtrl instead of "Unknown" object


bigpilot:
The current implementation generates an Unknown object for the RichTextCtrl even though wxWidgets XRC implementation seems to have support for a RichTextCtrl (src/xrc/xh_richtext.cpp)

Code needs to be changed in plugins/additional/additional.cpp (RichTextCtrlComponent::ExportToXrc())

sodevel:
I happily merge a PR that implements this

guysoft:
Hey, has this been released? The last release v3.9.0 seems to be from 2018.

bigpilot:
It's probably present in the master branch. Build from source if you want it.

PR #619, opened 2020-07-09
Add wxRichTextCtrl XRC and value property support

Kvaz1r:
Fix: #618 

I've not added [property maxlength](https://docs.wxwidgets.org/trunk/overview_xrcformat.html#xrc_wxrichtextctrl) because currently [does nothing](https://docs.wxwidgets.org/trunk/classwx_rich_text_ctrl.html#a3ac4c0377a88864857e678e2bf5a7f14).

### DIFF
--- a/plugins/additional/additional.cpp
+++ b/plugins/additional/additional.cpp
@@ -229,12 +229,20 @@ class RichTextCtrlComponent : public ComponentBase
 {
 public:
 	wxObject* Create(IObject* obj, wxObject* parent) override {
+
+		wxString text = obj->GetPropertyAsString(_("value"));
+
 		wxRichTextCtrl* richText = new wxRichTextCtrl( 	(wxWindow*)parent,
 															wxID_ANY,
-															wxEmptyString,
+															text,
 															obj->GetPropertyAsPoint(_("pos")),
 															obj->GetPropertyAsSize(_("size")),
 															obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
+
+		if (!text.empty())
+		{
+			return richText;
+		}
 
 		wxFont textFont = wxFont(12, wxFONTFAMILY_ROMAN, wxFONTSTYLE_NORMAL, wxFONTWEIGHT_NORMAL);
 		wxFont boldFont = wxFont(12, wxFONTFAMILY_ROMAN, wxFONTSTYLE_NORMAL, wxFONTWEIGHT_BOLD);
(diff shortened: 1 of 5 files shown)


plugins/additional/additional.cpp
```diff
@@ -229,13 +229,21 @@ class RichTextCtrlComponent : public ComponentBase
...
+
 		wxRichTextCtrl* richText = new wxRichTextCtrl( 	(wxWindow*)parent,
 															wxID_ANY,
-															wxEmptyString,
+															text,
 															obj->GetPropertyAsPoint(_("pos")),
 															obj->GetPropertyAsSize(_("size")),
 															obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
 
+		if(!text.IsEmpty())
```
sodevel, line 242: Please don't use wxWidgets API if there is std::string API available, here i would prefer to use `empty()`. Also add a space between the `if` and the parentheses.

plugins/additional/additional.cpp
```diff
@@ -370,14 +378,18 @@ class RichTextCtrlComponent : public ComponentBase
...
 
 	ticpp::Element* ExportToXrc(IObject* obj) override {
-		ObjectToXrcFilter xrc(obj, _("unknown"), obj->GetPropertyAsString(_("name")));
-		//xrc.AddWindowProperties();
+		ObjectToXrcFilter xrc(obj, _("wxRichTextCtrl"), obj->GetPropertyAsString(_("name")));
+		xrc.AddWindowProperties();
+		if(!obj->GetPropertyAsString(_("value")).IsEmpty())
+		{
+			xrc.AddProperty(_("value"), _("value"), XRC_TYPE_TEXT);
+		}
```
sodevel, line 386: Better add this property always, even if it's empty, other elements do this as well

plugins/additional/additional.cpp
```diff
@@ -370,14 +378,18 @@ class RichTextCtrlComponent : public ComponentBase
...
 		return xrc.GetXrcObject();
 	}
 
 	ticpp::Element* ImportFromXrc(ticpp::Element* xrcObj) override {
 		XrcToXfbFilter filter(xrcObj, _("wxRichTextCtrl"));
-		filter.AddWindowProperties();
```
sodevel, line 380: That looks like a mistake, the standard properties should also be handled

Kvaz1r: Definitely mistake, fixed. Also updated corresponding to comments above.

plugins/additional/additional.cpp
```diff
@@ -229,13 +229,21 @@ class RichTextCtrlComponent : public ComponentBase
...
+
 		wxRichTextCtrl* richText = new wxRichTextCtrl( 	(wxWindow*)parent,
 															wxID_ANY,
-															wxEmptyString,
+															text,
 															obj->GetPropertyAsPoint(_("pos")),
 															obj->GetPropertyAsSize(_("size")),
 															obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
 
+		if( !text.empty() )
```
sodevel, line 242: There must have been a slight misunderstanding here, i didn't mean to add padding inside the parentheses (like is seen in old code and shouldn't be used in new code) but just add a space after the if, so this should look like this:
```
if (!text.empty())
```

Kvaz1r: Right, old style a bit confused me. Fixed.

Issue #355, opened 2018-02-10
wxGrid: support wxGRID_AUTOSIZE to autosize column label height


Zenju:
It seems wxFormBuilder only allows positive integers for "col_label_size" of a wxGrid. 
Thous it's not possible to enter "-1" which is wxGRID_AUTOSIZE.

The hacky workaround currently is to enter "4294967295", which will be internally converted to "-1":
![grid](https://user-images.githubusercontent.com/7656530/36060656-0740b148-0e4e-11e8-9753-be46cca758f1.PNG)

Wish: Support negative values for "col_label_size".

jmoraleda:
I am working on allowing `col_label_size` and `col_row_size` to be left empty (so they take their default values) and I will add this case too.

jmoraleda:
The pull request mentioned above completely addresses this feature request:

After applying the PR, It is possible to specify _column label height_ and _row label width_ as:

* numerical value in pixels (just like before)
* option `wxGRID_AUTOSIZE`
* leave blank (the new default) which will lets wxWidgets use default values for these properties

PR #626, opened 2020-07-27
wxGrid new feature plus bug fix in PT_EDIT_OPTION

jmoraleda:
This PR is for a new feature: To allow `col_label_size` and `row_label_size` in `wxGrid` control to take on their default values, as well as the option `wxGRID_AUTOSIZE`. This fixes issue #355 

To add this feature I converted both properties from "uint" to "editoption".

--------------------------

When I made this conversion I discovered a bug in how "editoption" (`PT_EDIT_OPTION`) was implemented: When the edit value is not one of the valid options, the result should be parsed from the string not 0 as it was. I fixed this bug in a separate, completely independent commit for clarity and to make the code review easier. 

To fix this bug in a modular way, I added a new method `bool    IsMacro(const wxString &str )` to `TypeConv`  that returns whether the current string value of a property corresponds to a registered macro or not.

### DIFF
--- a/output/plugins/additional/xml/data.cppcode
+++ b/output/plugins/additional/xml/data.cppcode
@@ -265,9 +265,12 @@ Written by
 			@}
 			#nl $name->EnableDragColMove( $drag_col_move );
 			#nl $name->EnableDragColSize( $drag_col_size );
-			#nl $name->SetColLabelSize( $col_label_size );
 			#foreach $col_label_values
 			@{ $name->SetColLabelValue( #npred, #pred ); @}
+			#ifnotnull $col_label_size
+			@{
+				#nl $name->SetColLabelSize( $col_label_size );
+			@}
 			#nl $name->SetColLabelAlignment( $col_label_horiz_alignment, $col_label_vert_alignment );
 			#nl
 			#nl // Rows
@@ -278,9 +281,12 @@ Written by
 				#nl $name->AutoSizeRows();
 			@}
 			#nl $name->EnableDragRowSize( $drag_row_size );
-			#nl $name->SetRowLabelSize( $row_label_size );
 			#foreach $row_label_values
 			@{ $name->SetRowLabelValue( #npred, #pred ); @}
+			#ifnotnull $row_label_size
+			@{
+				#nl $name->SetRowLabelSize( $row_label_size );
+			@}
 			#nl $name->SetRowLabelAlignment( $row_label_horiz_alignment, $row_label_vert_alignment );
 			#nl
 			#nl // Label Appearance
--- a/output/plugins/additional/xml/data.luacode
+++ b/output/plugins/additional/xml/data.luacode
@@ -169,9 +169,12 @@ Lua code generation written by
 				#nl #utbl$name:AutoSizeColumns()
 			@}
 			#nl #utbl$name:EnableDragColSize( $drag_col_size )
-			#nl #utbl$name:SetColLabelSize( $col_label_size )
 			#foreach $col_label_values
 			@{ #utbl$name:SetColLabelValue( #npred, #pred ) @}
+			#ifnotnull col_label_size
+			@{
+				#nl #utbl$name:SetColLabelSize( $col_label_size )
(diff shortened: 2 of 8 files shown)


sodevel:
Splitting both changes into two commits was good, splitting them into two PR's would have been better because a PR should be focused on one topic.

Changing the property type has some drawbacks. The value in question is actually an integer value but that value can not only be specified by an integral value but one symbolic name. I remember i had the same problem when i added better support for wxDataView columns, they accepted a symbolic name for the value -1. There i simply kept the integer type because the control prevents the user from entering arbitrary strings, if he wants the special value he has to enter its integral value. The situation here is similar with one more difficulty, you need to specify the value as "not set". This is a big problem in wxFB, it is not possible for certain values to (re-)set them to "not set" (looking at you fonts and colors). Changing the type to editoption adds this possibility, but now it is also possible to enter random strings and they get accepted without error.

Sadly, currently there is no other option available to realize that feature easily so im going to accept this.

About the `Property::GetValueAsInteger()` i think this can be solved in a more simple way. Unless i miss something all the types PT_OPTION, PT_EDIT_OPTION and PT_MACRO seem to be used only for symbolic names of integral values, instead of performing a double dictionary lookup `TypeConv::GetMacroValue(const wxString&)` could do the integer conversion itself if the macro is unknown.

jmoraleda:
Thank you. I have made the change suggested: 

Instead of adding a new method `bool IsMacro(const wxString &str )` to `TypeConv` and checking for the macro existence when retrieving the value of the `Property`, `TypeConv::GetMacroValue(const wxString&)`  does the integer conversion itself if the macro is unknown.

sodevel:
Thanks for the changes, there is one small thing left regarding code style. New code shouldn't use this whitespace padding around parentheses, can you remove that please? Only in the cpp-files though, the code templates continue to use this padding because it affects the generated code.

jmoraleda:
Thank you. I removed the whitespace around the parentheses in `typeconv.cpp` to match the surrounding code style. I left the other files untouched after checking that the code style in my additions matched the surrounding code.

sodevel:
Thanks for the update, but despite that your new code fits the surrounding one, i like to continue the current project policy to write new code only in current style, even if that doesn't match the context code (i know this can be quite difficult sometimes and look very weird, im open for discussion in extreme cases).

So it would be great if you could update `additional.cpp` and `pythoncg.cpp` as well. Hopefully in the future there will be a better autoformatter available that can be applied unconditionally, currently clang-format does some very strange things to fix too long lines and i don't know a better (free) solution that works on all three platforms easily.

jmoraleda:
I think all code in a file should be consistent. I am happy to add a separate commit to change the formatting in `additional.cpp` and `pythoncg.cpp` by removing  the whitespace around all the parentheses in these files if this suggestion makes sense to you.

sodevel:
I do not want to introduce unnecessary formatting only changes. The final format that will be used in the future to clean up the codebase is not specified yet so it's wasted time to reformat code now. If you insist on following the surrounding style, well, i guess i have to live with that, i don't want to waste too much time on formatting issues but concentrate to get a new release finished soon.

jmoraleda:
I agree with you about wanting a consistent style through the complete code base, but until that happens, which perhaps may take a while, I still think keeping files self consistent keeps them more maintainable than migrating piece meal a few lines at a time.

I added a commit removing whitespace inside parentheses everywhere in the two files I modified so they still are self consistent and comply with the new guidelines at least with respect to space and parentheses. Personally, I would be equally happy without this last commit, and keeping those files self consistent in the previous style until the time when all the code base is reformatted in one large sweeping commit, so it is completely up to you whether you want to include this last commit when you do the merge. 

Thank you.

sodevel:
Your last commit does exactly what i wanted to prevent, really huge formatting-only changes. I would also be happy to drop that last commit, unfortunately i don't think i can do this easily through GitHub and i don't know what i mess up if i merge this manually. I think it's easier you force push without that commit so i can simply merge this PR.

PR #630, opened 2020-08-11
Add support for EVT_LIST_ITEM_CHECKED and EVT_LIST_ITEM_UNCHECKED t…

jmoraleda:
This pull request is for a new feature: It adds support for `EVT_LIST_ITEM_CHECKED` and `EVT_LIST_ITEM_UNCHECKED` to `ListCtrl`.

Background:

This events were added to ListCtrl in 3.1 (https://docs.wxwidgets.org/trunk/classwx_list_ctrl.html) which was released in February 2016 (https://en.wikipedia.org/wiki/WxWidgets)

### DIFF
--- a/output/plugins/common/xml/common.xml
+++ b/output/plugins/common/xml/common.xml
@@ -245,6 +245,8 @@ Written by
     <event name="OnListColDragging" class="wxListEvent" help="The divider between columns is being dragged."/>
     <event name="OnListColEndDrag" class="wxListEvent" help="A column has been resized by the user."/>
     <event name="OnListCacheHint" class="wxListEvent" help="Prepare cache for a virtual list control."/>
+    <event name="OnListItemChecked" class="wxListEvent" help="The item has been checked."/>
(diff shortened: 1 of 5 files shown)


output/plugins/common/xml/common.xml
```diff
@@ -245,6 +245,8 @@ Written by
     <event name="OnListColDragging" class="wxListEvent" help="The divider between columns is being dragged."/>
     <event name="OnListColEndDrag" class="wxListEvent" help="A column has been resized by the user."/>
     <event name="OnListCacheHint" class="wxListEvent" help="Prepare cache for a virtual list control."/>
+    <event name="OnListItemChecked" class="wxListEvent" help="The item has been checked. "/>
```
sodevel, line 248: Just one very small issue, there is trailing whitespace at the help string, would be great if you could remove it.

jmoraleda: Thank you. Done.

PR #631, opened 2020-08-14
CMake build on MSYS2 and Linux

michallukowski:
- My implementation is based on the cmake_build branch
- CMake build works on MSYS2 and Linux
- Instruction how to build and run added to README.md
- No hard coded path
- Default installation to cmake_install directory
- The installation copies all libraries needed to run wxFormBuilder
- Build and install path added to .gitignore


sodevel:
First of all thanks for the taking the time working on the CMake task. I am still quite unfamilar with CMake but i think i can say there is improvement over the previous attempt to get CMake integrated. However i also see issues that are causing trouble. I really want to use CMake as the only build system for wxFB, but for this it have to be usable for all supported platforms. This would be the following:

* Windows
  * MinGW (MSYS2)
  * MSVC (not supported by current systems)
* Linux
  * GCC flatpak
  * GCC native
* MacOS
  * GCC AppPackage

Your current approach seems to be limited to MSYS2 and Linux, this is kind of unfortunate because currently MacOS requires the most hacks to build at all. In the following review im adding some comments about what i think is problematic.

A general remark, PR's do trigger the CI. This is a great way to actually test if CMake does cover the current platforms, you can change the CI configuration to use CMake instead of the PreMake and Meson approach.

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+set(CPACK_PACKAGE_VERSION_MAJOR 3)
+set(CPACK_PACKAGE_VERSION_MINOR 9)
+set(CPACK_PACKAGE_VERSION_PATCH 0)
+set(CPACK_RESOURCE_FILE_LICENSE ${PROJECT_SOURCE_DIR}/COPYING)
+set(CPACK_WIX_UPGRADE_GUID B5AAB525-34D5-4B12-B6F2-136C8ED890C9)
+
+include(CPack)
+
+set(CMAKE_CXX_FLAGS "-Wno-deprecated-declarations ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS_DEBUG "-D__WXFB_DEBUG__ ${CMAKE_CXX_FLAGS_DEBUG}")
```
sodevel, line 31: This is a gcc specific option and gcc specific syntax to set a define, this won't work with non-gcc compilers (MSVC and clang)

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+set(INSTALL_ROOT_DIR "cmake_install")
+
+if(MSYS)
+    set(INSTALL_BIN_DIR ".")
+    set(INSTALL_SHARE_DIR ".")
+    set(INSTALL_LIB_DIR ".")
+elseif(UNIX)
+    set(INSTALL_BIN_DIR "bin")
+    set(INSTALL_SHARE_DIR "share/wxformbuilder")
+    set(INSTALL_LIB_DIR "lib/wxformbuilder")
```
sodevel, line 42: Especially setting the lib directory is very problematic. Right now wxFB has issues to load its libraries because different linux distributions love to use different directories. CentOS based ones use lib64, Ubuntu based ones use an architecture depended subdirectory

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+endif()
+
+set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/${INSTALL_ROOT_DIR})
+
+set(CMAKE_INSTALL_SYSTEM_RUNTIME_DESTINATION ${INSTALL_BIN_DIR})
+include(InstallRequiredSystemLibraries)
+
+include_directories(subprojects/ticpp)
+add_definitions(-DTIXML_USE_TICPP)
```
sodevel, line 51: This is deprecated even in your targeted CMake version

plugins/additional/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+    install(FILES
+        ${CMAKE_CURRENT_BINARY_DIR}/libadditional${CMAKE_SHARED_LIBRARY_SUFFIX}
+        DESTINATION ${INSTALL_LIB_DIR}/plugins/additional/
+    )
+elseif(UNIX)
+    install(FILES
+        ${CMAKE_CURRENT_BINARY_DIR}/libadditional${CMAKE_SHARED_LIBRARY_SUFFIX}
+        DESTINATION ${INSTALL_LIB_DIR}
+    )
+endif()
```
sodevel, line 18: Why differentiate between the platforms, can't the installation directory propertly set for each platform?
Why is the output library name hardcoded, can't it be queried from the target?

sodevel (changes requested):
In general there is only the distinction between MSYS and UNIX, this is not sufficient.
The plugin integration could be extracted into a function, there is a lot of copy-paste.

CMakeLists.txt
michallukowski: I have added an if condition.

CMakeLists.txt
michallukowski: Problem not yet resolved. I added these paths based on where wxFormBuilder looks for libraries.

CMakeLists.txt
michallukowski: I have got rid of redundant code.

plugins/additional/CMakeLists.txt
michallukowski: Code replaced by function. I added these paths based on where wxFormBuilder looks for plugins.

michallukowski:
I prepared a cmake build for MSYS2 and Linux as I only use these platforms. I can't add support for other platforms. Someone else is needed for this.

Issue #625, opened 2020-07-18
Dark mode issues on macOS


LoganDark:
Now, I recognize that nobody has gotten wxFormBuilder to build and run on a recent version of macOS in a long time (#525). However, I've managed to get mine to work simply by building wxWidgets from source and using a hack to make wxFormBuilder find the library it wants (https://github.com/wxFormBuilder/wxFormBuilder/issues/525#issuecomment-660439943).

Here's the current state of the UI:
![image](https://user-images.githubusercontent.com/4723091/87859050-3cb95180-c8e7-11ea-83ec-d7935f2610af.png)

There are a few issues right off the bat:
- In the left tree view, selected items turn white when the tree view is not in focus (when it's in focus things look fine)
- AUI panel titles have a bright grey background
- Inactive AUI tab colors are grey and not very visible
- Window titlebars are grey and the title is not visible (in this example, it's set to "Login")
- Background colors in the property inspector are slightly obnoxious

I'm not sure how much of this can be fixed, and I've _also_ been told that none of the main developers have a Mac themselves. However, these are all quite simple tweaks and since C++ is one of my native languages I'd be happy to open a pull request if there's a need for it.

LoganDark:
Note that the tree view thing manifests itself **exactly the same** in really old versions of wxFormBuilder that don't support the dark theme.

![image](https://user-images.githubusercontent.com/4723091/87859304-054ba480-c8e9-11ea-8f29-ccb9d64324b5.png)

It looks like the tree view hasn't been changed at all in this time. A proper fix would also fix the text color on light theme.

Kvaz1r:
I seen recently several changes to AUI dark mode under OSX in wxWidgets repo, for example this one - [https://github.com/wxWidgets/wxWidgets/pull/1955](Fixes for AUI in dark mode under macOS) 
Did you build wxFormBuilder with those changes?

LoganDark:
I downloaded the source from wxWidgets' website, if you want me to try building master I can

Kvaz1r:
Yes, it would be great - maybe some of issue already was fixed and some maybe related to wxWidgets itself.

sodevel:
> There are a few issues right off the bat:
> 
>     * In the left tree view, selected items turn white when the tree view is not in focus (when it's in focus things look fine)
> 
>     * AUI panel titles have a bright grey background
> 
>     * Inactive AUI tab colors are grey and not very visible

[...]

PR #632, opened 2020-08-14
Change workspace background colour in dark mode.

michallukowski:
When program is in dark mode the background is almost white. Changing wxSYS_COLOUR_APPWORKSPACE to wxSYS_COLOUR_WINDOW makes the program look better.

### DIFF
--- a/src/rad/designer/visualeditor.cpp
+++ b/src/rad/designer/visualeditor.cpp
@@ -75,7 +75,7 @@ m_stopModifiedEvent( false )
 	}
 	else
 	{
-		SetOwnBackgroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_APPWORKSPACE).ChangeLightness(120));
+		SetOwnBackgroundColour(wxSystemSettings::GetColour(wxSYS_COLOUR_WINDOW).ChangeLightness(120));
 	}
 
 	SetScrollRate(5, 5);


sodevel:
I'm quite reluctant about merging this, dark mode color schemes seem to be a quite tricky thing. Take a look at #625, the current background seems to work quite well for MacOS and one CentOS 7 theme, but this also depends on which wxWidgets version is used. Can you give some details on which system it doesn't work well, maybe together with a screenshot? OS version, used theme and wxWidgets version/commit.

michallukowski:
I'm using Linux Manjaro with Gnome. wxWidgets 3.0.5. 
Here is a few screenshots how it looks before and after changes. Every theme I checked looks similar.

Before:

before_dark

After:

after_dark

Before:

before_dark2

After:

after_dark2

sodevel:
I tested your changes on CentOS 7 and Fedora 32 and for some strange reason i don't get any difference at all. This kind of makes sense because the color choosers inside wxFB show the same color for both elements. However i am using a much newer wxWidgets version than you https://github.com/wxWidgets/wxWidgets/commit/6a2005945aafcd9e3e81661f815a7a2064a96cc7 and this version fixed a lot of dark mode issues for me.

I am not sure what the meaning of those colors is exactly and if the present use is actually correct, but the components shown on that background usually have the WINDOW color so i wanted to use another one for the surrounding to prevent the lightness change beeing the only difference to create contrast. APPWORKSPACE sounds reasonable for me, kind of strange that in my environments it seems to be identical to WINDOW, luckily the lightness change works quite well.

If a newer wxWidgets version fixes the issue i would prefer to keep the current state.

PR #639, opened 2020-09-21
Add read_only support to StyledTextCtrl

jmoraleda:
This pull request adds support for the `read_only` property of widget `StyledTextCtrl` to all generators that currently support the StyledTextCtrl widget (currently cpp and python).

### DIFF
--- a/output/plugins/additional/xml/additional.cppcode
+++ b/output/plugins/additional/xml/additional.cppcode
@@ -396,6 +396,7 @@ Written by
 			#nl $name->SetViewWhiteSpace( $view_whitespace );
 			#nl $name->SetMarginWidth( 2, 0 );
 			#nl $name->SetIndentationGuides( $indentation_guides );
+			#nl $name->SetReadOnly( $read_only );
 			#ifequal $folding "1"
 			@{
 				#nl $name->SetMarginType( 1, wxSTC_MARGIN_SYMBOL );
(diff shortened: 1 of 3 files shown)


sodevel:
It's kind of unfortunate that this setting gets only applied in two of the code templates and not even mentioned in the others. I don't remember exactly how this was handled previously by the code templates, but i would suggest to generate the method calls as comment, maybe with an additional comment line above that mentions that this is currently not supported. This way it is easier to spot these locations later if these methods get supported by the other languages as well. Even if i'm quite sure that these two language bindings are kind of dead.

jmoraleda:
Thank you. A comment to indicate that `StyledTextCtrl` is not support already exists in the other two language bindings. In particular https://github.com/wxFormBuilder/wxFormBuilder/blob/master/output/plugins/additional/xml/additional.luacode#L252 and https://github.com/wxFormBuilder/wxFormBuilder/blob/master/output/plugins/additional/xml/additional.phpcode#L271

jmoraleda:
@sodevel What are your thoughts on this?

sodevel:
Sorry for the late reply, i am very busy currently.

For some strange reasons i thought this control is supported by all languages but does not support the read only flag in every language. Later i realized the control is completely unsupported and the generators already output that comment, so my last comment is kind of void.

PR #642, opened 2020-09-24
Adding components to the project form the Components menu or by searching in the dialog window

michallukowski:
Add: Components menu based on the Component Palette
Add: Dialog to find component by typing its name. Calling the dialog window from "Components menu" or shortcut Ctrl+Shift+f
Add: Additional file to meson.build

Components menu:

wxformbuilder_components_menu_01

Finding dialog:

wxformbuilder_components_menu_02


sodevel:
Sorry, im currently very busy and can't do a detailed review.

This looks like a nice addition, some small remarks:
- I would put the menu right to the edit menu, it's kind of an edit function and shouldn't be so far away from it
- Is it possible to use the present bitmaps also in the menu? Might make it easier to spot the desired control, but also might overload the menu.
- This search dialog looks a bit weird, i would expect some insert or cancel button there. How do you actually insert a selected element from this dialog?

I will take a closer look at this when i have more time.

michallukowski:
Thanks for your reply and suggestions. I have added icons to the components menu and move close to the edit menu:

wxformbuilder_components_menu_03

The search dialog is very simple now. Its operation is as follows:

- Press "Shift+Ctrl+f" or select from the menu to show the search dialog
- Just type whatever you want to find and matching components will appear below
- Press Tab to switch to the component list or use mouse
- Use Key Up, Key Down or mouse to select appropriate component
- Use Enter key, Space key or double click to insert the component to the project
- Use Esc key or click X button to cancel 

If I find some free time, I will add title and buttons to the dialog.

In the future, we can expand the dialog with more functionality (component description, images how component looks like)

michallukowski:
I have added window title, buttons Cancel and Insert. Insert button is active after selecting component from list. 

wxformbuilder_dialogfindcomponent_01 wxformbuilder_dialogfindcomponent

PR #650, opened 2020-11-10
Added support to Erlang language (wxErlang)

Micheus:
I hope you can accept the addition of this language - even it not being "class oriented" as the others.
Because of that some particular tag are hardcode handled.

I included an readme file to be used by erlang users in order to understand some workarounds related to events and for others who would like to setup the configuration files in the future (this way no one needs to start from scratch as I have to) .

I made a build available [here](https://github.com/Micheus/wxFormBuilder/releases/tag/v3.9.0-wxErlang) in case someone wants to check how it is working.

Some screenshots:
**Project settings for Erlang**
wxProjectProperties

**Source generation kind: Create window only**
curve-tools-prototype2

**Source generation kind: Full code prototype**
curve-tools-prototype3

**AUI management - creation**
wxAUI-managed-code1

**AUI management - control handled**
wxAUI-managed-ctrls

**Events - Handling the different kinds**
wxFormEvents


Kvaz1r:
I checked your version from release and it's wonderful. I noticed some minor things (not sure did you fix it in current version or not, unfortunately I haven't time to re-checking it now, but they were minor, copy-paste typos mostly).
It's really great work and I hope this PR will be merged, thank you once again, you made Erlang available for all who prefer RAD tools.

Micheus:
> I checked your version from release and it's wonderful. I noticed some minor things (not sure did you fix it in current version or not, unfortunately I haven't time to re-checking it now, but they were minor, copy-paste typos mostly).

The source code is rebased with the current commits in the master.
The executable I made available doesn't include the commits since October, but as you said they are minor updates.

sodevel:
Thank you for this PR, unfortunately due to severe time contraints i can't take a closer look at it in the near future.

By looking at the screenshots i have two questions so far. Do the generated files always contain that full copyright notice? You can only change the name with the project properties? If this is the case, is there any strong reasoning why the generated code falls under the Apache License? So far all generators don't apply any license. If this is just for personal preference i would suggest you make the full license block a project property instead of just the name so that users can inject the license of their choice, or none.

Are there now some magic codes for event handler methods? I don't know any Erlang so what are they needed for? Can't this be controlled by a global property instead like for the C++ case where you can decide if you want just declared handlers, pure virtual ones or implemented virtual ones? Is it really necessary to change this on a by handler case?

Micheus:
**Question block 1**
Q: _Do the generated files always contain that full copyright notice?_
A: Its common session in any Erlang source code we see.

Q: _You can only change the name with the project properties?_
A: Yes

Q: _If this is the case, is there any strong reasoning why the generated code falls under the Apache License?_
A: It was chosen by convenience. In the Erlang sources we can usually find the _Apache Licence_ or _Erlang Public License_.

S: _I would suggest you make the full license block a project property._
A: I can do that. 

**Question block 2**
Q: _Are there now some magic codes for event handler methods?_
A: Yes?!
By considering the ***.<_language_**>**code** files are specific for the language being handled, I didn't see any problem to handle the event handlers in a different way. So, the only way I found it would work to make available the synch and asynch events in Erlang since it doesn't handle classes or virtual classes (or something like that).

Each option is connected to the control in a different way. Naming a event handler means we can only use the callback events assigned to a function. That is why I'm first checking for the [_callback_, _skip_ and _assync_] before parse the event handler in order to allow the user to make use of these options.

For instance, the _**paint**_ event must be handled in callbacks which will requires a declaration like this:
`xPanel:connect(Canvas, paint, [callback]),` (check the note [here](https://github.com/erlang-labs/wx/blob/master/examples/demo/ex_canvas_paint.erl#L99) in an wxErlang sample file)

Q: _Is it really necessary to change this on a by handler case?_
A: Yes, since a project can use more than one way to connect events as in the sample in the image I attached.

Kvaz1r:
Any progress with review of the PR?

Micheus:
> Any progress with review of the PR?

Hi **Kvaz1r**,
I was waiting for any counter arguments to my answers. In the mean time I went back to work on other stuff.
As it didn't come I ended up by forgetting about some changes I can do.

Micheus:
@sodevel, I just added a new commit [867957f](https://github.com/wxFormBuilder/wxFormBuilder/pull/650/commits/867957feeef800874ff653c2feb37a81dbd300f4) to apply the suggestion you did (questions in block 1).

For the questions in block 2, you didn't replied, so I assume it may be valid my arguments for them.
Just completing, I also made that kind of customisation because I saw something similar in Lua code, otherwise I would not try this implementation since this tool would not be so useful to Erlang users.

If you consider this PR should not be added to the project I hope you don't mind if I keep this branch in my repository.

Anyway, thanks for your attention.

sodevel:
Thank you for the changes, unfortunately im still in the process of catching up things and this is kind of a big addition and will need some more time. In your repository you can do whatever you want :wink:

Micheus:
> 1. The generated code changes depending on the selected element in the object tree. This is a huge blocker, you can see the effect by using the test project. Open the Erlang tab and select a child element of a different toplevel element, the generated code changes. I found the location in the code that is responsible for this, but i don't know why this is done and how it can be fixed.

@sodevel , I'm not sure about what is wrong in this item.
For me, it seems its behave is the same as for the other languages. By selection any element on the tree the code presented is the one for its toplvel form  (MyFrame1, MyPanel1 or MyDialog1).
In case the user is selecting the project, then all the three form's code is shown together.

How is it supposed to work? If you could make it more explicit to me I would to appreciate. 


> 2.The "local event table code" is very hard to understand, does very strange things and currently causes crashes. Just use one of these special event handler names outlined above and you get a crash.

I believe they are hard to understand because Erlang works different with wx events ([wxEvtHandler](https://www.erlang.org/doc/man/wxevthandler#connect-3)), do strange things maybe because the source code is different for each of choice.

By using your 'test' file, in the MyFrame1 I set  the properties as bellow (the three reserved words and one event identifier):
```
OnActivateApp: async
OnClose: OnClose
OnMaximize: skip
OnMove: callback
```
and I got this into the source code, plus the proper events handle and function (which is correct):
```
	%% Connect Events
	wxFrame:connect(MyFrame1, activate_app),
	wxFrame:connect(MyFrame1, close_window, [{callback, fun onclose/2}]),
	wxFrame:connect(MyFrame1, maximize, [{skip, true}]),
	wxFrame:connect(MyFrame1, move, [callback]),
```
but, no crash. Where the crash is happening for you?

p.s. I'm using this changed code you worked on (not my old one).

sodevel:
> @sodevel , I'm not sure about what is wrong in this item. For me, it seems its behave is the same as for the other languages. By selection any element on the tree the code presented is the one for its toplvel form (MyFrame1, MyPanel1 or MyDialog1). In case the user is selecting the project, then all the three form's code is shown together.
> 
> How is it supposed to work? If you could make it more explicit to me I would to appreciate.

Strange, for some reason i cannot reproduce the output of the Erlang panel that i have described. Maybe i was confused that the `init()` function of the full code prototype changed, apparently it can only create the first top level element and not the others. Anyway, the code changes still happen. Using the previous test file, select `MyProject1` in the Object Tree and generate the code. Rename the generated file. Select `MyDialog1` in the Object Tree and generate the code. Compare both generated files and observe that in the `MyDialog1` case suddenly all child elements use `MyDialog1` as parent instead of the top level element that contains them.


> I believe they are hard to understand because Erlang works different with wx events ([wxEvtHandler](https://www.erlang.org/doc/man/wxevthandler#connect-3)), do strange things maybe because the source code is different for each of choice.

Thats not what i meant, i was talking about the implementation inside the code generator that realizes these different event handler types. It's hard to understand what is done there, what the components of the local event table actually are and how it is created and why it is deleted during the generation process.

On a second note, the `wxErlang_README.md` file mostly describes how the template system works in general and what controls are not supported by Erlang. There is currently no documentation about what special event handler keywords exist except the screenshots in this PR.


> By using your 'test' file, in the MyFrame1 I set the properties as bellow (the three reserved words and one event identifier):
> 
> ```
> OnActivateApp: async
> OnClose: OnClose
> OnMaximize: skip
> OnMove: callback
> ```
> [...]
> 
> 
> but, no crash. Where the crash is happening for you?

It is actually a debug assertion of the MSVC runtime that terminates the application, it is caused by incrementing an invalidated iterator, see the attached backtrace.

```
wxFormBuilder.exe!std::_Tree_const_iterator<std::_Tree_val<std::_Tree_simple_types<std::pair<wxString const ,std::map<wxString,wxString,std::less<wxString>,std::allocator<std::pair<wxString const ,wxString>>>>>>>::operator++() Line 198 (c:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xtree:198)
wxFormBuilder.exe!std::_Tree_iterator<std::_Tree_val<std::_Tree_simple_types<std::pair<wxString const ,std::map<wxString,wxString,std::less<wxString>,std::allocator<std::pair<wxString const ,wxString>>>>>>>::operator++() Line 288 (c:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xtree:288)
wxFormBuilder.exe!ErlangCodeGenerator::ClearLocalEventTable(const wxString & strEvtName) Line 684 (c:\git\xprivate\wxformbuilder\src\codegen\erlangcg.cpp:684)
wxFormBuilder.exe!ErlangCodeGenerator::GenEventHeaders(std::shared_ptr<ObjectBase> project, std::shared_ptr<CodeInfo> code_info) Line 1048 (c:\git\xprivate\wxformbuilder\src\codegen\erlangcg.cpp:1048)
wxFormBuilder.exe!ErlangCodeGenerator::GenerateCode(std::shared_ptr<ObjectBase> project) Line 664 (c:\git\xprivate\wxformbuilder\src\codegen\erlangcg.cpp:664)
wxFormBuilder.exe!ErlangPanel::OnCodeGeneration(wxFBEvent & event) Line 313 (c:\git\xprivate\wxformbuilder\src\rad\erlangpanel\erlangpanel.cpp:313)
wxFormBuilder.exe!ErlangPanel::OnEventHandlerModified(wxFBEventHandlerEvent & event) Line 177 (c:\git\xprivate\wxformbuilder\src\rad\erlangpanel\erlangpanel.cpp:177)
wxmsw316ud_vc.dll!00007ff8da4af067() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da4aefb9() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da634818() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da632fbf() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da6353d6() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da48de74() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da633c7e() (Unknown Source:0)
wxmsw316ud_vc.dll!00007ff8da633ba6() (Unknown Source:0)
wxFormBuilder.exe!ApplicationData::NotifyEvent(wxFBEvent & event, bool forcedelayed) Line 3025 (c:\git\xprivate\wxformbuilder\src\rad\appdata.cpp:3025)
wxFormBuilder.exe!ApplicationData::NotifyEventHandlerModified(std::shared_ptr<Event> evtHandler) Line 3087 (c:\git\xprivate\wxformbuilder\src\rad\appdata.cpp:3087)
wxFormBuilder.exe!ApplicationData::ModifyEventHandler(std::shared_ptr<Event> evt, wxString value) Line 1216 (c:\git\xprivate\wxformbuilder\src\rad\appdata.cpp:1216)
wxFormBuilder.exe!ObjectInspector::OnEventGridChanged(wxPropertyGridEvent & event) Line 911 (c:\git\xprivate\wxformbuilder\src\rad\inspector\objinspect.cpp:911)
wxmsw316ud_vc.dll!00007ff8da4af067() (Unknown Source:0)
```

Micheus:
@sodevel, I didn't forget about your comments and to make the possible changes you have required. I'm just very busy with personal stuff at this time. I hope to get back to it in January.

Micheus:
Hi @sodevel 
I finely got some time to get back this project and update it to be inline with all the changes you did this year.

Unfortunately, I didn't find the place I have to add the plugin's settings to include all '*.erlangcode' in the build/install script. I found the function on cmake script that should run this copy process, but not where the list is initialized.

Can you please help me with that?

sodevel:
Generator languages are defined with the variable `WXFB_GENERATOR_LANGUAGES` in the main `CMakeLists.txt` file in the root directory. You only have to add the base name of the language there, you don't need to add the code template filenames anywhere, their names are calculated from the plugin data. Additional resources that are required by the main application can be added to the various variables defined at the beginning of the `CMakeLists.txt` file in the `src` directory.

I did all this already in commit dacf4d5e74d82b8698d1c8304ebbf0c5d84d0d4f. But i didn't adapt this branch to the new project layout. Basically, you need to move your files to the proper locations, you can take the other plugins as reference. Unless i miss something, there should be no need to modify the CMake files any further than done in the mentioned commit.

Micheus:
> I did all this already in commit [dacf4d5](https://github.com/wxFormBuilder/wxFormBuilder/commit/dacf4d5e74d82b8698d1c8304ebbf0c5d84d0d4f).

`set(WXFB_GENERATOR_LANGUAGES cpp python lua php erlang CACHE INTERNAL "")`
This line was the missing one I was looking for changes. Thanks.

Back to validate the rebase.

p.s.: I just noticed your read.me point the run code to (on Windows - MSYS2): `_install/bin/wxFormBuilder`, but no `bin` folder was created and the executable is under `_install` folder.

sodevel:
> p.s.: I just noticed your read.me point the run code to (on Windows - MSYS2): `_install/bin/wxFormBuilder`, but no `bin` folder was created and the executable is under `_install` folder.

Fixed.

> Also, the `meson.build` file listed in the conflicts report bellow cannot be found at `master` branch.

This is because i removed the Meson build files, but since you made changes to this file, this causes a conflict. Just remove that file as well to solve it.

Micheus:
Hi @sodevel
I just saw there were a lot happening on this project since the last time I looked at it. I just get retired and now I intend to reserve Friday afternoon to work on this branch. So, I hope to update the old code and fix the things you asked for before in the next months.

PR #668, opened 2021-04-17
Adds wxSimpleBook XRC export code.

### LINKED ISSUE #666: wxSimplebook written as "unknown" class to XRC.
Using XRC generation, a wxSimplebook control is written as an "unknown" class rather than "wxSimplebook.

I get the following as an example:
`<object class="unknown" name="m_parameterBook" />`

This means more unnecessary work to either fix the generated XRC or to load the controls correctly from "unknown".

This could of course be something I'm doing wrong, but I'm not quite sure what it could be. I've tested this on a snapshot build of master as well, not just the release builds.

OmniBlade:
Implements the missing XRC import/export code in plugins/containers/container.cpp.
Fixes #666

### DIFF
--- a/plugins/containers/containers.cpp
+++ b/plugins/containers/containers.cpp
@@ -880,6 +880,18 @@ public:
 			obj->GetPropertyAsSize(_("size")),
 			obj->GetPropertyAsInteger(_("window_style")));
 	}
+    
+    ticpp::Element* ExportToXrc(IObject* obj) override {
+		ObjectToXrcFilter xrc(obj, _("wxSimplebook"), obj->GetPropertyAsString(_("name")));
+		xrc.AddWindowProperties();
+		return xrc.GetXrcObject();
+	}
+
+	ticpp::Element* ImportFromXrc(ticpp::Element* xrcObj) override {
+		XrcToXfbFilter filter(xrcObj, _("wxSimplebook"));
+		filter.AddWindowProperties();
+		return filter.GetXfbObject();
+	}
 };
 
 class SimplebookPageComponent : public ComponentBase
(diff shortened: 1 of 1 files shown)


plugins/containers/containers.cpp
```diff
@@ -892,6 +904,21 @@ class SimplebookPageComponent : public ComponentBase
 	void OnSelected(wxObject* wxobject) override {
 		BookUtils::OnSelected< wxSimplebook >( wxobject, GetManager() );
 	}
+    
+    ticpp::Element* ExportToXrc(IObject* obj) override {
+		ObjectToXrcFilter xrc(obj, _("simplebookpage"));
+		xrc.AddProperty(_("label"),_("label"),XRC_TYPE_TEXT);
+		xrc.AddProperty(_("select"),_("selected"),XRC_TYPE_BOOL);
```
sodevel, line 911: Space after comma missing

plugins/containers/containers.cpp
```diff
@@ -892,6 +904,21 @@ class SimplebookPageComponent : public ComponentBase
...
+		return xrc.GetXrcObject();
+	}
+
+	ticpp::Element* ImportFromXrc(ticpp::Element* xrcObj) override {
+		XrcToXfbFilter filter(xrcObj, _("simplebookpage"));
+		filter.AddWindowProperties();
+		filter.AddProperty(_("label"),_("label"),XRC_TYPE_TEXT);
+		filter.AddProperty(_("selected"),_("select"),XRC_TYPE_BOOL);
```
sodevel, line 919: Space after comma missing

sodevel (changes requested):
This is looking good, there are just some minor formatting issues, it would be nice if you could fix them.

OmniBlade:
Looks like the spacing issues is prevelent in the file as I just copy pasted those sections and modified them :D I've fixed them now, but left the rest of the file as is so the PR just reflects the functional change.

sodevel:
Formatting is kind of a mess in the code base, the long term goal is to implement a format specification to allow automatic formatting of the code. The difficulty is to find a free solution that works on all three platforms easily, so far i only know about clang-format, but i am quite unhappy with some of the results it produces.

Anyway, thanks for the PR, i fixed the introduced whitespace errors myself.

PR #670, opened 2021-04-30
Added .editorconfig

redtide:
For those like me that uses it other than clang-format.
I noticed late that you are using tab for indentation? I can force push to change spaces to tabs but old code still have spaces, so before to change here I would like to be sure which should be used.
FWIW space with a size of 4 is the most used and recommended.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = 1
+tab_width = 4
+trim_trailing_whitespace = 1


sodevel:
The code style is currently a mess and i haven't found a good solution for this yet. My goal is to use clang-format, but last time i tried it still produced a result i wasn't satisfied with. For now im reluctant to endorse any specific style, just blend into the surrounding style. For completely new code use spaces and Java like brace placement.

redtide:
I use my own style but with a [project](https://github.com/sfztools/sfizz/blob/develop/.clang-format) I follow we use [WebKit](https://webkit.org/code-style-guidelines/), which looks also the most used.

redtide:
Coding without this and/or clang-format is going to be annoying, would be great if we agree to use again 4 spaces as it was the original code, but following the [Stroustrup coding style](https://www.stroustrup.com/Programming/PPP-style.pdf) instead the original wxWidgets coding convention, which IIRC was removed in some commit, and merge this in, clang-format adaption included.

sodevel:
I know the current situation is quite unpleasant but unfortunately not so easy to change. The current codebase mixes multiple styles and smart editors (looking at you Visual Studio) tend to reformat the whole file once they see a .clang-format file even if you only change a line. This kind of requires to reformat the whole code base, this was rejected previously for reasons and i still have my objections against this currently. I reformatted the XML and Code Template files before and still regret the choice to keep the TABs in the latter, but this introduced the least amount of changes.

In my opinion Clang-Format is the way to go because it's free and works on all the three platforms, but i did not and still do not agree how it wraps long lines. I can live with most of the two coding styles you mentioned, there are some small details i do disagree with them, e.g. unbraced one-liners. I was working on a project with Clang-Format recently and managed to create a configuration that fulfilled almost all of my requirements, i'll try to adapt it for wxFB and post it here soon for discussion.

PR #672, opened 2021-05-01
Move premake scripts from build directory

redtide:
This is a first step for a reorganization of the files to permit to add a `scripts/cmake` modules directory.
I would like also to reorganize other scripts and resources in a way to avoid to mix up build scripts, resources and build objects.
The `build` directory should not be used for build tools and scripts.

The plan would be also to remove the need of the `output` directory and let the build tool (Premake, CMake or Meson) to copy all files in the appropriated build destination directory in a usual way, using a `resources` source directory in the root.

Would be also appropriated to move also the `install` directory into `scripts`.

I think the current state is unnecessary complicated and difficult to understand for new comers.

I can take care to all of this myself if nobody have time to work on it ATM.


sodevel:
In general a cleanup of the structure is appreciated, but care has to be taken to not break something by accident. Most probably some very old elements can be removed though because they are not used anymore.

I'm not sure if it's feasable to remove the output directory yet, should be quite some work to adapt Premake, and this would be kind of wasted because that should be replaced in the future anyway. About the install directory, at least some of the resources do not belong into the scripts directory, not sure though whats a better place to keep them.

.gitignore
```diff
@@ -1,6 +1,6 @@
-build/???/
-build/premake/**/bin/
-build/premake/**/obj/
+/*build*/
```
sodevel, line 1: Why the wildcards? Wouldn't be a simple /build/ enough?

redtide: it matches any of `/_build` `/builddebug` etc

redtide:
> In general a cleanup of the structure is appreciated, but care has to be taken to not break something by accident. Most probably some very old elements can be removed though because they are not used anymore.

of course, same for the install directory, things need to be tested before as I did.

> I'm not sure if it's feasable to remove the output directory yet, should be quite some work to adapt Premake, and this would be kind of wasted because that should be replaced in the future anyway. About the install directory, at least some of the resources do not belong into the scripts directory, not sure though whats a better place to keep them.

The `output` directory is against the usual developer choice where to build the code, which is usually the `build` directory and it was removed already in the `cmake_build` branch, thing that I avoid to do before adding the CMake scripts, so meanwhile I move Premake scripts in the same place where to add later the cmake ones and so remove completely the `build` directory to be used for compiling, not for tools. So the structure would be:

```
/resources
/scripts
- cmake
- install
- premake
/src
```

PR #674, opened 2021-05-06
CMake build configuration

redtide:
This (third) alternative CMake build configuration is currently used by myself to code on wxFB, it is tested with Archlinux with wx 3.0.5, MSYS2 and Visual Studio 2019.

I add it as draft PR because I'm still updating it by force pull commits (this is why I marked not editable), it lacks of minor features but it's not yet tested/working on macOS and used in any CI.
The `Utils` module is just for testing, will not part of it and the minimum version required should be less recent, will change it later once all main features will be in.

I took some parts from the `cmake_build` branch (MinGW dependencies).
It uses CMake modules scripts, so it doesn't invasively infest the source code tree, just a single CMakeFiles.txt in the root directory and modules in a `scripts/cmake` folder.

Opened mainly for feedback, requests/suggestions discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,47 @@
+cmake_minimum_required(VERSION 3.15)
+project(wxFormBuilder
+    VERSION 3.9.1
+    LANGUAGES CXX C
+    DESCRIPTION "An Open Source GUI Builder for wxWidgets"
+    HOMEPAGE_URL "http://wxformbuilder.org"
+)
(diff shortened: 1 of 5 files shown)


sodevel:
Since i still know almost nothing about CMake i can't say much about it from a technical perspective, i can only speak as a noob user here and what i would expect. It would be nice if i could click on Open Folder inside VSCode, select Visual Studio (on Windows) or GCC / Clang (on Linux) as toolkit, press Compile and receive a binary i can simply execute 😃

redtide:
Working on my own fork now and going to delete the branch, closing.

Issue #665, opened 2021-03-10
gmake fails on macOS Big Sur 11.2.3


patrickleevers:
Linking wxFormBuilder
    Running post-build commands
    sh ../../../install/macosx/postbuild.sh -c release
    * Collecting dependencies.......
    /!\ WARNING : Cannot resolve path '/usr/lib/libSystem.B.dylib'
    /!\ WARNING : Library libSystem.B.dylib has an incomplete name (location unknown)
    Please specify now the directory where this library can be found (or write 'quit' to abort):

The problem is explained in this thread (today!):
https://developer.apple.com/forums/thread/655588

Huskyforce:
Same Problem here

patrickleevers:
It's looking as if macos support for this project has been abandoned.… A great pity, as Qt's is slowing too and I was hoping for better from wxWidgets.

sodevel:
This project is not part of wxWidgets nor supported by it in any way. Currently there is only one active maintainer and that one has currently no time available. Additionally that maintainer does not have any macOS experience nor access to any macOS based system beside the CI system employed in this project.

Since Apple seems to go the Microsoft way and make their own standards, instead of ranting it would be more productive if someone with macOS experience and/or access to macOS would try to fix this issue and open a PR here.

CricYoung:
[Here](https://wiki.rocrail.net/doku.php?id=cmac-en)
Here is a solution may be can help.
Support Apple M1 chip.
I succeeded after doing exactly what he say.

vinidiktov:
> [Here](https://wiki.rocrail.net/doku.php?id=cmac-en)
> Here is a solution may be can help.
> Support Apple M1 chip.
> I succeeded after doing exactly what he say.

I've tried both the official instructions and the solution that you linked to.

For me, in both cases the resulting app crashes the same way as is described in https://github.com/wxFormBuilder/wxFormBuilder/issues/677

sodevel:
Does this problem still occur when using the new wxwidgets package from homebrew outlined in the updated install instructions?

PR #675, opened 2021-05-06
Fix AUI notebooks tab text color with dark theme on wx < 3.1.4

### LINKED ISSUE #625: Dark mode issues on macOS

redtide:
Quick backport of 3.1.4 code to fix the issue on wx 3.0 based systems.
Fixes #625.

Apparently this doesn't affect Windows because it seems that dark mode, introduced in Windows 10, needs to be enabled from the application (or in wx libraries) and the API are still in a work in progress. (see [this post](https://forums.wxwidgets.org/viewtopic.php?t=47110#p205959) on wx forum).

Note that I tested it only on Linux, haven't the chance to get even `master` working on my macOS BigSur VM: it builds and installs with dylibbundler postbuild scripts, but the resulting app won't run.

### DIFF
--- a/src/rad/auitabart.cpp
+++ b/src/rad/auitabart.cpp
@@ -1,0 +1,511 @@
+#include "auitabart.h"
+
+#ifndef WX_PRECOMP
+    #include <wx/dc.h>
+    #include <wx/dcclient.h>
+    #include <wx/settings.h>
+    #include <wx/bitmap.h>
+#endif
+
+#include <wx/aui/auibook.h>
+#include <wx/renderer.h>
+
+#if defined( __WXMAC__ )
+ static const unsigned char close_bits[]={
+     0xFF, 0xFF, 0xFF, 0xFF, 0x0F, 0xFE, 0x03, 0xF8, 0x01, 0xF0, 0x19, 0xF3,
+     0xB8, 0xE3, 0xF0, 0xE1, 0xE0, 0xE0, 0xF0, 0xE1, 0xB8, 0xE3, 0x19, 0xF3,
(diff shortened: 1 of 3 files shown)


sodevel:
I don't think this PR is such a good idea, basically you copy the custom render code from wxWidgets. Now if wxWidgets changes this code for some reason the visual appearance can divert and you have to synchronize the changes again. And all this just to fix something that fixes itself by upgrading wxWidgets.

MacOS builds are broken for recent versions, i don't think i can do much about this myself because all i have is the CI and it was too cumbersome to figure out how to make the build work with Premake v5 already. But maybe #665 can help you to get it working in your VM.

redtide:
Thank you for the tip, closing.

PR #678, opened 2021-05-27
Final installation steps on linux x86_64

sosie-js:
Tested on Debian 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ ninja -C _build install
 ./_install/bin/wxformbuilder
 ```
 
+Extra steps to finish install on linux x86-64 (Debian):
+
+```sh
+sudo cp -rd _install/* /usr; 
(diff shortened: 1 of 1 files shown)


sodevel:
You are mixing up two different things here. You perform a system wide installation from the local installation described in the current installation manual. This is NOT how to install wxFB system wide, specify the proper prefix to the install command instead.

The second thing is a workaround for the current defect that wxFB cannot load plugins on platforms that use architecture specific library subdirectories. This affects most prominently Debian based systems, to some extent also RedHat based systems are affected but wxFB contains an ugly hack to work there (the lib64 symlink). This is actually a defect of the wxWidgets library which simply cannot resolve such directories, apparently no one fixes this issue.

Adding that symlink for a local install is an acceptable solution but i cannot recommend doing this for a system wide installation.

Issue #679, opened 2021-06-01
AUI toolbar dropdown menus misaligned


eric135:
When clicking a dropdown menu generated for an AUI toolbar with wxFormBuilder 3.9.0 and wxWidgets 3.1.5, the dropdown menu is misaligned compared to the tool it was dropped down from. I have noticed that the alignment is only slightly off for the leftmost toolbar. However, if I add a second toolbar to the right, the menu now appears outside of the window. I have produced this issue on both Windows 10 and Ubuntu 20.10 and have confirmed that this happens in a minimal app with only two AUI toolbars. Here are screenshots of this behavior:

![Screenshot 2021-05-31 205746](https://user-images.githubusercontent.com/6131490/120264721-3871fa80-c253-11eb-885a-68b8bc4c9890.png)

![Screenshot 2021-05-31 205818](https://user-images.githubusercontent.com/6131490/120264729-3c9e1800-c253-11eb-814c-4ce9086c821d.png)

Kvaz1r:
Could you provide code (as .fpb file) for reproducing the behaviour?

eric135:
Sure, here's the .fbp file (in a .zip because GitHub wouldn't let me upload with it called *.fbp).

[wxFormBuilderTest.fbp.zip](https://github.com/wxFormBuilder/wxFormBuilder/files/6574583/wxFormBuilderTest.fbp.zip)

Kvaz1r:
Ok, reproduced. 
Replace this:
`m_auiToolBar2->PopupMenu( m_menu2, pt );`
with this:
`PopupMenu( m_menu2, pt );`
for every toolbar. 

I'm going to open PR with fix (just update [this](https://github.com/wxFormBuilder/wxFormBuilder/blob/0eccdbc5256b58cd553f68b6ad70b7ca0355da3e/output/plugins/common/xml/menutoolbar.cppcode#L107) line)

PR #680, opened 2021-06-01
Fix misalign for auitoolbar dropdown menu

Kvaz1r:
Fix: #679


sodevel:
This change looks correct, but what about the other supported languages? Aren't they affected by the bug as well?

Kvaz1r:
Updated for python and php. I don't know for sure about Lua because it has same syntax for all tools.

sodevel:
It looks like the Lua generator uses the parent element to convert the coordinates to but also uses the parent element to show the popup menu, so this might work. Anyway, since i also don't know the Lua details, we should keep it as is and wait for complaints. I think C++ and Python are the most (and only?) used generators so we should focus on these.

PR #681, opened 2021-06-06
Fix various typos

luzpaz:
Found via `codespell -q 3 -L iif,msdos,pevent`


sodevel:
Thats an interesting tool, thanks for the fixes.

Issue #685, opened 2021-06-25
Android dialogs fonts need to be set before objects created


jongough:
Using wxFormBuilder to provide the base class for all display objects in a multi-platform development it has worked well. But have now added android as one of the supported platforms and issues have arisen with the c++ code that is generated. On android the default font of an object, i.e. dialog box, needs to be set before any child objects, i.e. static text, are created. If this is not done the use of 'SetFont(....)' will set the font of the child objects but the size of these objects will be calculated using the original font size when they were created. 

The wxWidgets manual says, in a round about way, that a dialog box needs to be instantiated first, then SetFont needs to be called to set the default font, then the display objects need to be added to the dialog box. In the current implementation, 3.9.0, this is not possible as the class definition does the child object creation.

I have done a test by changing the code that creates the c++ class to have the instantiation only create the class, and have added a method, create(), which can be used by a child class to create all the objects. This allows for the setting of the default font to something else than the system font.

The original code:
```
ODPropertiesDialogDef::ODPropertiesDialogDef( wxWindow* parent, wxWindowID id, const wxString& title, const wxPoint& pos, const wxSize& size, long style ) : wxDialog( parent, id, title, pos, size, style )
{
	this->SetSizeHints( wxDefaultSize, wxDefaultSize );
```
The modified code:
```
ODPropertiesDialogDef::ODPropertiesDialogDef( wxWindow* parent, wxWindowID id, const wxString& title, const wxPoint& pos, const wxSize& size, long style ) : wxDialog( parent, id, title, pos, size, style )
{}
void ODPropertiesDialogDef::Create()
{
	this->SetSizeHints( wxDefaultSize, wxDefaultSize );

```
There needs to be some changes to the wxFormBuilder GUI to allow the user to specify that the 'Create()' method be used or not. There are two possible solutions, either the base class will by default call the 'Create' method and if requested by the user not do so, or generate the extra 3 lines when asked to.

I do not know how to change the GUI to reflect this change or I would submit a pull request.

jongough:
Pull request, #686, generated. New option for wxWindow, create_children, on by default so class calls Create() method. If turned off the derived class will need to call this method to create all the children.

sodevel:
wxWidgets offers two ways to create widgets, either use the constructor with arguments to fully create the widget, or use the default constructor to only create the C++ part and use the Create() method (which always has arguments) to complete the creation. wxFB is using the former method, using the latter is not feasable because all code templates need to be changed.

[...]

PR #686, opened 2021-06-25
Add ability to delay creation of children of a dialog to allow settin…

jongough:
…g defaults prior to creation. This is needed for Android where if the defaults are not set, i.e. fonts, the size of the object is not correctly drawn.


jongough:
Addresses issue: [685](https://github.com/wxFormBuilder/wxFormBuilder/issues/685#issue-929709554)

output/xml/default.xml
```diff
@@ -132,6 +132,7 @@
     <property name="context_menu" type="bool" help="Generates event handler for displaying of menu assigned to this widgets as a context menu.">1</property>
     <property name="context_help" type="wxString_i18n" help="Attach context-sensitive help to the window. Note: The Project's &quot;help_provider&quot; property must be set for context-sensitive help to work."/>
     <property name="enabled" type="bool" help="Enable or disable the window for user input. Note that when a parent window is disabled, all of its children are disabled as well and they are reenabled again when the parent is.">1</property>
+    <property name="create_children" type="bool" help="Automatically create children when class instansiated. c++ only">1</property>
```
sodevel, line 135: Ignoring the typographic errors i don't think this property belongs here. This form of two-step initialization can only be used for toplevel widgets, because these are the only ones the user can use directly. If this gets enabled for child elements it will break the code because the generated code doesn't know anything about this initialization type and will always use the one-step initialization.

Why is this limited to C++ only? Shouldn't it be available for the other languages too?

sdk/plugin_interface/xrcconv.cpp
```diff
@@ -413,7 +413,10 @@ void ObjectToXrcFilter::AddWindowProperties()
 	if (!m_obj->IsNull(_("enabled")) && m_obj->GetPropertyAsInteger(_("enabled")) == 0) {
 		AddProperty(_("enabled"), _("enabled"), XRC_TYPE_BOOL);
 	}
-	if ( !m_obj->IsNull( _( "focused" ) ) )
+	if (!m_obj->IsNull(_("create_children")) && m_obj->GetPropertyAsInteger(_("create_children")) == 0) {
+        AddProperty(_("create_children"), _("create_children"), XRC_TYPE_BOOL);
+    }
```
sodevel, line 418: This is useless because the XRC "language" doesn't define this property, the XRC handlers shipped with wxWidgets simply won't read this property

src/codegen/cppcg.cpp
```diff
@@ -1226,6 +1227,7 @@ void CppCodeGenerator::GenClassDeclaration(PObjectBase class_obj, bool use_enum,
 	// protected
 	m_header->WriteLn( wxT( "protected:" ) );
 	m_header->Indent();
+    m_header->WriteLn( wxString::Format( wxT( "void Create();\n" ) ) );
```
sodevel, line 1230: Like mentioned in the associated issue, this name is kind of misleading because it is already used by wxWidgets for two-step creation

src/codegen/cppcg.cpp
```diff
@@ -1585,8 +1587,20 @@ void CppCodeGenerator::GenConstructor(PObjectBase class_obj, const EventVector &
...
+    if ( class_obj->GetPropertyAsInteger(wxT("create_children")) != 0 ) {
+        m_source->WriteLn( wxT( "{" ) );
+        m_source->Indent();
+        m_source->WriteLn( wxT( "Create();" ) );
+        m_source->Unindent();
+        m_source->WriteLn( wxT( "}" ) );
+    } else {
+        m_source->WriteLn( wxT( "{" ) );
+        m_source->WriteLn( wxT( "}" ) );
+    }
```
sodevel, line 1599: This looks like a pretty hacky attempt to get the second method defined, i didn't take a detailed look at the present code but i think it would be a cleaner way to create the method definition in a dedicated method

src/codegen/cppcg.cpp
```diff
@@ -1585,8 +1587,20 @@ void CppCodeGenerator::GenConstructor(PObjectBase class_obj, const EventVector &
...
+    wxString className = class_obj->GetPropertyAsString( wxT( "name" ) );
+    m_source->WriteLn( wxString::Format( wxT( "void %s::Create()" ), className.c_str(), className.c_str() ) );
+    m_source->WriteLn( wxT( "{" ) );
+    m_source->Indent();
```
sodevel, line 1603: This will break the subclass feature and this will only work for toplevel elements

sodevel (commented):
In general this solution is not clean and robust enough to merge. I think a proper implementation is quite some work and adds too much complexity for not that much value. I think arbitrary code injection into the generated code offers more value with less work.

src/codegen/cppcg.cpp
jongough: Not knowing any of the background to wxFormBuilder this was an attempt to make the generated code work for Android builds. Currently whilst the dialogs do build and display the result is quite often unreadable. If the name 'Create' is a problem that is easily changed. The default behaviour does not affect any current objects, the user will have to select to not auto create the children.

There is no current method of setting the default fonts to use other than with the 'SetFont' method after the dialog is instantiated. The current wxFormBuilder process does not allow for the default font to be set before the children are created. In Android the only way to then change the font is to cycle through all children and set their font explicitly.

I only use C++ so I only made the changes to affect the C++ part of the system. I do not know enough about the other languages to even start to do that.

If this solution is totally unacceptable then there needs to be a method of setting the default font before the children are created. I am not sure what you mean by 'code injection' or the use of a plugin, so I am in the dark about this possible solution.

jongough:
The pull has been updated with changes that make this only apply to C++ dialog boxes. No other form elements are affected. The method name has been changed from 'Create()' to 'CreateChildren' to avoid overloading an existing function. The default action will be to create the new method and call it from the instantiation. 

Are there any other issues with this process?

sodevel:
Im sorry, but these changes make it even worse.

Now highly generic code checks for the magic value ``Dialog`` to change its behavior, this is just not acceptable. Also why only Dialog, what about the other toplevel elements Frame, Panel and Wizard?

I just pushed an update that contains the CustomCode component, this allows you to change the font programmatically before child elements get added so they inherit the proper font. If the font is "static" and known at compile time you could also just set the font property of the toplevel element, this also gets set before child elements get added. Any of these two approaches should solve your issue.

jongough:
The fonts are user pickable at runtime, so they are not static and are not known at compile time.

Where is your push, I cannot see it in this git repository? I would like to have a look at it and do some testing.

jongough:
The change made to include Custom Code does not help. Whilst code can be added it is constrained to the class being generated as there is no way to get to the child class or its parents which is needed to set a font. The code is in a place that it could be done, but the class definition would have to be changed to have a public member that can be called with a variable number of attributes such that these could be used to provide the information needed to do the work, in this case a wxFont object would need to be sent to the Dialog class for the custom code to use.

sodevel:
In my use-case i can access a global object to get the required data, if in your use-case this is not possible or desired the current options aren't sufficient. I think in that case two-step creation is the only viable clean solution and i think one of the purposes it exists (unless setting a font on such a default created object is not possible).

The reason why i didn't like your first approach was it was too broad, your second attempt is too narrow and adds logic and even code generation into a part where it doesn't belong. Code generation is controlled by the various code templates of the different languages (``*.<language>code`` files). The code generators basically execute these templates.

Two-Step creation makes only sense for toplevel elements, usually these are the only ones the user can interface directly. There are only 4 toplevel elements that could make use of two-step creation, all have their own code templates (part of the ``forms`` plugin), it should be possible to implement that feature for them without any code changes but only updating the code templates.

jongough:
Having looked through the template files there is currently no way to separate the instantiation from the create children. It would appear that the template language does not allow this to occur. Is there any documentation on this language or how  to implement new methods within the generated class?

sodevel:
I'm afraid there is no such documentation, iirc #650 contains some description of the template language, other than that you have to look at the source in ``src/codegen/``.

Actually there is a certain order in which these templates get executed, this can be used create additional methods in the proper templates. I did a test in the branch two-step-creation (279666a5685acfc13e3006d49a79cda25c60e6d7) for wxFrame, seems to work but currently breaks the Generate Inherited Class feature because this requires the constructor with arguments that i dropped for simplicity.

This approach should allow to implement that feature for all languages, however i really only know C++ well, so for me this is also quite hard to do for the other languages. In my opinion PHP and Lua can be neglected, Python however seems to be quite popular and wxFB seems to be one of the few (the only one?) wxWidgets GUI builder for that language, so this language should support this feature.

jongough:
I have updated the code to use templates. Is this better?

output/plugins/forms/xml/forms.cppcode
```diff
@@ -407,7 +407,26 @@ Written by
 			#ifequal $aui_managed "1"
 			@{ @#include &lt;wx/aui/aui.h&gt; @}
 		</template>
-		<template name="settings">
+        <template name="code_init">
```
sodevel, line 410: The name doesn't match the present naming style wich uses the name components in reversed order

output/plugins/forms/xml/forms.cppcode
```diff
@@ -407,7 +407,26 @@ Written by
...
+            #ifnotnull $create_children
```
sodevel, line 411: This test is pretty pointless because a boolean property can never be null

output/plugins/forms/xml/forms.cppcode
sodevel, line 410: This name doesn't match the present naming style, the name components are in reversed order

output/plugins/forms/xml/forms.cppcode
sodevel, line 411: This test is useless, a boolean property can never be null

output/plugins/forms/xml/forms.cppcode
```diff
@@ -407,7 +407,26 @@ Written by
...
+            @{
+            #ifequal $create_children "0"
+            @{ { #nl
+            } #nl #nl
+            void $name::Init()
+            @}
+            @}
```
sodevel, line 418: This is hard to read because of no indentation

sodevel (commented):
Yes, this is a better approach. Since you also rely on the workflow of the settings template getting inserted into whatever method that was declared before, it is not really necessary to introduce new code templates. The constructor templates get executed right before the settings template, the code could be added there.

Is there any reason why you are still moving the initialization code off the parameterized constructor into an own method? The delayed execution of the wxFB generated code can be achieved by using two-step creation. I think supporting this standard approach of wxWidgets is of greater value than just refactoring the one-step creation approach.

I have updated the two-step-creation branch (934f9c2ffc2de79ede390017b194fb0d97e35003) to support two-step creation for each toplevel element, currently only for C++ though. Please take a look if this is working for you.

output/plugins/forms/xml/forms.cppcode
jongough: What if it doesn't exist as in when this change is first implemented?

output/plugins/forms/xml/forms.cppcode
jongough: That will be fixed, I was just trying stuff out to see if it would work.

output/plugins/forms/xml/forms.cppcode
sodevel: The properties are part of the object model that gets defined in the XML files. If the XML file defines the property, it is available with a valid value, no matter if your project file contains a value for it or not. If your code template tries to access a property that is not defined in the XML files, this is a programmer error and results into strange errors. These errors can't be prevented by such a check because the check itself already causes an error.

output/plugins/forms/xml/forms.cppcode
jongough: <snip>
> Is there any reason why you are still moving the initialization code off the parameterized constructor into an own method? The delayed execution of the wxFB generated code can be achieved by using two-step creation. I think supporting this standard approach of wxWidgets is of greater value than just refactoring the one-step creation approach.
<snip>
 
I want to have the class created normally with all the parameters passed, but I don't want any children created before setting the font. If I don't do this explicit calls will have to be made to functions to do what can be done by just passing parameters at instantiation. To me it is cleaner. It also allows other calls to be made if needed before creating extra child objects.

output/plugins/forms/xml/forms.cppcode
jongough: Latest push should have addressed the issues you raised (thanks).

output/plugins/forms/xml/forms.cppcode
sodevel: But this is what two-step creation enables you to do. Keep in mind you don't use the generated class directly, instead you derive a class from it and use that one. With two-step creation you can define a parameterized constructor in your derived class but still default create the (generated) base class. Before you call the Create() method you can call other methods first, e.g. change the font. Granted, if you don't call Create() in that constructor, you need to save some arguments to use them later.

I just merged the two-step creation feature because it is a nice addition to wxFB. There you can see that no additional code templates are required (minus the one that gets executed after event handler connection, but this might be useful for future enhancements too). Also this is kind of the state i would a PR expect to reach to be able to merge it.

If you still need that Init() approach i would suggest to make it an alternative to two-step creation because i don't think it's easy to support both together nor does this make sense. For that purpose i would suggest to replace the `two_step_creation` boolean property with an option property, maybe `delayed_construction`, that has the values `<empty>` for one-step creation only, `two-step` for two-step creation and `init` for your proposed Init() method.

sodevel:
Please try to avoid pulling in commits from master if you do not really need them, this just makes it harder to read the PR. A better alternative would be to rebase the PR against master and force-push instead.

output/plugins/forms/xml/forms.cppcode
jongough: This has the problem that for every change, i.e. new methods that need to be called (style is not the only one that may be needed, and android proved that) that may be needed between instantiation of the class and creating all the children you will have to make extra changes to the Create class. My version achieves the same but allows the developer to change the parts they require before building all the children and therefore if there are other objects/settings/etc. that need changing they can be. The resultant code is almost the same but with the Create you have lock in and inbuilt maintenance.

jongough: This change also does not change the default style before creating child objects. The whole point of this change was to allow changes to be made to the default settings prior to them being instantiated. In Android if you do not set the default style to what is required before creating children the only way to change the style is to update each child element individually. So the change I put is has the desired effect, so I will update my pull request with your code so that it can be pulled in.

jongough:
This version includes two_step_creation and create_children. The are aimed at different results.

Create_children is to allow the full instantiation of the class but not to automatically create the children. This allows manipulation the defaults of the class before building children based on these defaults. This is currently required on android deploys or the changed defaults are only partially applied, i.e. changes to fonts are accomplished but the sizing of the children is based on the original fonts, so the display is poorly rendered.

output/plugins/forms/xml/forms.cppcode
sodevel: But this is exactly what two-step creation is made for. I don't know how i can explain the situation better, so im trying to show you two simple examples instead. Just to make sure, the usual process to use the wxFB generated class is not to use the class directly but inherit from it. With enabled two-step creation you have basically two choices how to modify the parent class before the generated code adds the components.

Example 1: In-Class modification inside derived class
```c++
// Generated by wxFB
class MyGeneratedDialog;

class MyDialog : public MyGeneratedDialog {
public:
   MyDialog(wxWindow* parent, wxWindowID id = wxID_ANY, const wxString& title = wxEmptyString, const wxPoint& pos = wxDefaultPosition, const wxSize& size = wxDefaultSize, long style = wxDEFAULT_DIALOG_STYLE) {
      SetFont(*wxITALIC_FONT);
      SetExtraStyle(wxWS_EX_PROCESS_IDLE);
      SetForegroundColour(*wxRED);

      Create(parent, id, title, pos, size, style);
   }
};

int main() {
   auto* dlg = new MyDialog(nullptr);

   return 0;
}
```

Example 2: External modification
```c++
// Generated by wxFB
class MyGeneratedDialog;

class MyDialog : public MyGeneratedDialog {};

int main() {
   auto* dlg = new MyDialog();
   dlg->SetFont(*wxITALIC_FONT);
   dlg->SetExtraStyle(wxWS_EX_PROCESS_IDLE);
   dlg->SetForegroundColour(*wxRED);

   dlg->Create(nullptr);

   return 0;
}
```

Why can't you use any of these approaches?

sodevel:
I don't know what happened here, but now there are 30 commits, 95% beeing picked from master, some even twice. I don't know what you did, but this is not a rebase. Also as far as i can see, you did not change anything in your code. As i said before in the long conversation of the first code snippet, two-step creation and your init method cannot be used together (very complicated to get the code templates right for every possible combination and also quite useless because both try to achieve a similar effect). If, after reading my examples above, you still think you need the init approach, please do what i suggested above: change the property type (i could do that for you if needed) and integrate your code templates in the present constructor templates (no need to add more templates, this requires to add them to the generators and templates of all the other languages).

If you don't know how to rebase this PR (there should be like 5 commits left if rebased, even less if you squash some of the work-in-progress ones), i'm not sure if i can rebase it for you.

jongough:
I have tried out your method and got it working. Thanks. So closing this pull request.

I did my normal process which does not cause problems in most repositories but I don't squash due to loss of information and potential issues with rebasing (which I have experienced and can take time to sort out).

When will your updated version be released to general use? I want to get away from using my personal copy so that others using my code can use wxFormBuilder without issues.

sodevel:
When (most of) the issues listed under the 3.10.0 milestone are resolved, unfortunately it seems this still will take some time.

PR #689, opened 2021-07-12
update flatpak platform to 20.08

jolz:
Because i use another distro, here is a proposal to remove dependency to "dnf".

I also propose to use "flatpak --user" environment, so I can build this on my work-system (without root).

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,19 @@
+# create a flatpak-build
+on: [push, pull_request]
+
+name: flatpak
+jobs:
+  flatpak:
+    name: "Flatpak"
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:gnome-40
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
+    - uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v3
+      with:
+        bundle: wxformbuilder.flatpak
+        manifest-path: install/linux/org.wxformbuilder.wxFormBuilder.json 
+        cache-key: flatpak-builder-${{ github.sha }}
+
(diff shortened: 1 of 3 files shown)


sodevel:
I see no problem updating the version, there is however a problem removing the `dnf` calls. This script is called by CI to setup the environment (iirc this is executed inside a docker container) to build the package, removing the calls will break the CI. Sadly, since the domain change at Travis the CI is currently gone and currently i don't have the time (and knowledge) to transform the configuration to GitHub Actions (which i prefer to use as only CI system). So right know i would like to keep this function present until the process of creating the flatpak works with GitGub Actions.

jolz:
I've just updated the script, so dnf-calls *should* work. Sorry to say - untested :-(

jolz:
Oups, I've just noticed, that my gitlab-ci tests shows up here.
The last commit seems to work with github-ci (it compiles), you may take this if you want. A deployment step is missing. 
Here is the source for my tests: https://github.com/marketplace/actions/flatpak-builder

Attention: This is one of my first pull requests and my very first github-action... Sorry...

sodevel:
Yes, this is how pull requests work, basically you pin your branch here, all its changes are visible here as well. If enabled, maintainers could even push to your branch to make some changes. The great thing about GitHub Actions is that they also work in forks out-of-the-box.

Actually, this is looking really nice, that flatpak-builder action even uploads the artifact to GitHub. I think a deployment step is not necessary currently, especially it should not deploy everything that gets build, only release builds should be deployed. I'm afraid it still needs some time for the next release.

The GitHub Action file looks ok, maybe a better name for some steps could be used, but this is not that important. These many work-in-progess commits don't look so nice, can you squash these commits into maybe two commits, one for the changes in the wxFB code and one for the GitHub Action, and force-push them?

jolz:
Sorry, been away for a while. 

Now the commits are squashed to 2 commits. Not shure, if you want to accept the github actions commit - it just checks, no deployment.

sodevel:
Thanks for the cleanup. As i said before, it is enough that the artifact gets uploaded to GitHub, it will be available there for 90 days. These automatically created builds should not be published, only builds of releases should be. Since releases are currently pretty rare, this can be done manually for now. Thanks for adding this additional CI action, this makes it easier to test the Linux builds.

Issue #712, opened 2021-12-28
wxTextCtrl add  drag file name 


wannianhong:
We often need to put the file name in the wxTextCtrl box 
wxTextCtrl need add property：SetDropTarget

code:
```
#pragma once
#include <wx/dnd.h>
#include <wx/textctrl.h>
#include <wx/arrstr.h>

 */
class CopyFilename : public wxFileDropTarget
{
private:
    wxTextCtrl* m_text;
public:
    CopyFilename(wxTextCtrl* text);
    virtual bool OnDropFiles(wxCoord x, wxCoord y, const wxArrayString& filenames);
    ~CopyFilename();
};
#include "CopyFilename.h"
CopyFilename::CopyFilename(wxTextCtrl* text) 
{
    m_text = text;
}

bool CopyFilename::OnDropFiles(wxCoord x, wxCoord y, const wxArrayString& filenames) 
{
    size_t nFiles = filenames.GetCount();
    wxString str;
    str.Printf(wxT("%d files dropped"), (int)nFiles);
    for (size_t n = 0; n < nFiles; n++) {
        m_text->AppendText(filenames[n]);
    }
    return true;
}
CopyFilename::~CopyFilename() {}

```

sodevel:
I'm sorry, but i'm afraid i don't understand the issue / request here.

wannianhong:
wxTextCtrl need add property：SetDropTarget
example:
m_obf_textCtrl->SetDropTarget(new CopyFilename(m_obf_textCtrl));

add SetDropTarget

wannianhong:
<img width="670" alt="edit" src="https://user-images.githubusercontent.com/59216544/147848846-915e5ac5-f342-4e64-a888-456919dd62b9.PNG">

wannianhong:
<img width="390" alt="e1" src="https://user-images.githubusercontent.com/59216544/147848866-fca3d05a-65ce-4b25-98dc-d6bbb111631a.PNG">
i can add property，but i don't know add new class CopyFilename

sodevel:
To keep the conversation in one place, i will answer in the linked PR.

PR #714, opened 2022-01-03
add wxTextCtrl property：SetDropTarget and msvc build

wannianhong:
add wxTextCtrl property：SetDropTarget,
add msvc build wxFormbuilder
fix https://github.com/wxFormBuilder/wxFormBuilder/issues/712


sodevel:
For a start, let me explain the task of wxFB. wxFB is used to create the visual appearance of application windows by creating the code to setup them. It is NOT designed to create the application logic of these windows, this is done externally by the user by deriving a class from the generated code.

This PR has two big issues. What you are trying here is to add some application logic that depends on application specific code. While wxFB currently already supports some very limited application logic (namely validators), this does not depend on application specific code but uses code supplied by wxWidgets.

This PR however uses code specific for your application (the drive_$name class), probably no one else beside you has any use for this. I cannot add this to wxFB, because if i would start doing so, other people might get the same desire to add their own code to wxFB and it would be cluttered with custom code fragments.

If you really want to define drag and drop handlers from inside wxFB the best option is to use the CustomCode component.

wannianhong:
CustomCode cann't  deriving a class

sodevel:
A CustomCode block gets inserted as-is approximately at the location in the generated code corresponding to the location in the object tree, e.g. if you place the element after the wxTextCtrl the code block gets inserted after the setup code of the wxTextCtrl.

Since you are using C++, you can define a local class inside a function, so you can basically create such a code block. But i don't see how this is useful, if you want to reuse this class, you have to define it everywhere else too. Regarding your example class from the issue, `CopyFilename`, this is better part of your project, maybe even part of a private library. All the CustomCode element would do in that case is to use the proper include directive and add the `SetDropTarget()` call.

wannianhong:
```
#include <wx/dnd.h>
// A drop target which accepts files (dragged from File Manager or Explorer)
class WXDLLIMPEXP_CORE wxFileDropTarget : public wxDropTarget
{
public:
    wxFileDropTarget();

    // parameters are the number of files and the array of file names
    virtual bool OnDropFiles(wxCoord x, wxCoord y,
                             const wxArrayString& filenames) = 0;

    virtual wxDragResult OnData(wxCoord x, wxCoord y, wxDragResult def) wxOVERRIDE;

private:
    wxDECLARE_NO_COPY_CLASS(wxFileDropTarget);
};
```
CopyFilename inherit wxFileDropTarget ，is not   private library

sodevel:
Yes, `wxFileDropTarget` is part of wxWidgets, `CopyFilename` is NOT. I have never used the wxWidgets drag and drop system so far, but from looking at the documentation, you always have to create an own class to be able to use it. Except you use the simpler `wxWindow::DragAcceptFiles(true)` and use `wxTextCtrl::OnDropFiles(wxDropFilesEvent &event)` inside the corresponding event handler, however the latter method seems to be available only on Windows, so you might to implement its function yourself on other platforms.

wannianhong:
Use DragAcceptFiles and wxEVT_DROP_FILES methods 
https://github.com/wxFormBuilder/wxFormBuilder/pull/719

wannianhong:
Use DragAcceptFiles and wxEVT_DROP_FILES methods
https://github.com/wxFormBuilder/wxFormBuilder/pull/719

PR #715, opened 2022-01-07
fix incorrect use of API `wxAuiTabContainer::AddPage`

GerHobbelt:
fix incorrect use of API `wxAuiTabContainer::AddPage` where `wxAuiTabContainer::InsertPage` was intended (4th argument used is the page index, not an image ID). This was uncovered by a wxWidget assertion (`wxFAIL_MSG`) firing in `wxWithImages::GetBitmapBundle(int iconIndex)` in an MSVC wxFormBuilder debug build on MSWindows.

(wxWidgets library version used during test: bleeding edge master branch ~ wxWidgets 3.1.6pre, but going through the wxWidgets history, this seems an older mistake as the 3.1.\* API `wxAuiTabContainer::AddPage()` has not been changed/augmented in any way that I could see would make the existing usage of that API correct. In other words: AFAICT this applies for building against **any wxWidgets 3.1.\* version**.)

### DIFF
--- a/src/rad/cpppanel/cpppanel.cpp
+++ b/src/rad/cpppanel/cpppanel.cpp
@@ -69,12 +69,12 @@ wxPanel( parent, id )
 
 	m_cppPanel = new CodeEditor( m_notebook, wxID_ANY);
 	InitStyledTextCtrl( m_cppPanel->GetTextCtrl() );
-	m_notebook->AddPage( m_cppPanel, wxT( "cpp" ), false, 0 );
+	m_notebook->InsertPage( 0, m_cppPanel, wxT( "cpp" ), false );
 	m_notebook->SetPageBitmap( 0, AppBitmaps::GetBitmap( wxT( "cpp" ), 16 ) );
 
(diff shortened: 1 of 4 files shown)


sodevel:
These asserts must have been added recently, i guess i should update my wxWidgets, thanks for spotting these.

I guess an even simpler fix would be to simply omit the parameter, the pages are added in the correct order. However, the following code that sets the images references the pages by index, so inserting the page at that index right before looks semantically more correct.

GerHobbelt:
That's why I chose to use Insert page: no hidden assumption about page
number anywhere in that code chunk.

And, yes, I think those asserts have indeed been added recently. I haven't
spent the effort to check exactly when (git bisect) so it *may* be that
they only exist in master branch. I did check for API changes as far back
as 3.1.1, which is why I wrote the pullreq subject line and commit message
like that.

Thanks for all your work on wxFormBuilder. It's really useful and allows me
to nitpick instead of having to build from scratch instead, which would be
a much higher cost in effort from my side for a similar end result.

PR #719, opened 2022-01-14
add wxTextCtrl property：DragAcceptFiles and msvc build

wannianhong:
I test windows and ubuntu environment, both support


wannianhong:
It has been modified

PR #720, opened 2022-01-14
Fix find wxWidgets 3.1 in MSYS2

michallukowski:
- Cmake cannot find wxWidgets version 3.1 in MSYS2 until "wxWidgets_CONFIG_EXECUTABLE" is set to "wx-config-3.1".

- set default installation prefix (MacOS check needed)


sodevel:
Since MSYS2 ships a wxWidgets 3.1 release the GitHub CI is using it and there was never a problem finding it. In what situations does this not work?

Looking at the wxWidgets Find Module i found this:
```cmake
    find_program(wxWidgets_CONFIG_EXECUTABLE
      NAMES
        $ENV{WX_CONFIG}
        wx-config
        wx-config-3.1 wx-config-3.0 wx-config-2.9 wx-config-2.8
        wxgtk3u-3.1-config wxgtk3u-3.0-config wxgtk2u-2.8-config
      DOC "Location of wxWidgets library configuration provider binary (wx-config)."
      ONLY_CMAKE_FIND_ROOT_PATH
      )
```

"Overriding" a (non documented) cache variable looks like a bad idea, setting that environment variable seems to be the better option.

michallukowski:
Perhaps this error occurs when we have installed two versions of wxWidgets. In MSYS2 '/mingw64/bin/wx-config' by default corresponds to version 3.0 and therefore we get an error during cmake configuration.

sodevel:
If thats the case it seems the best way to solve this is to set the environment variable WX_CONFIG to the required version, the Find Module should work without any modifications.

michallukowski:
You're right. I have corrected according to your advice.

CMakeLists.txt
```diff
@@ -87,6 +87,10 @@ set(WXFB_GENERATOR_LANGUAGES cpp python lua php CACHE INTERNAL "")
 
 if(NOT WXFB_WXWIDGETS_ENABLE_30)
   set(wxWidgetsVersion 3.1.2)
+
+  if(MSYS)
+    set(ENV{WX_CONFIG} "wx-config-3.1")
+  endif(MSYS)
```
sodevel, line 93: The purpose of these environment variables is that users can give information to CMake that is required to build on their machine without modifying the CMake files of the project. Now if you forcefully set this variable this isn't possible anymore. I don't think this variable should be set here, it should be set by external means (i am using MSVC and have setup scripts that prepare the environment, at least on this platform this is kind of a common approach).

If you really think this should be set in the project file, then please check if this environment variable is already set and don't touch it in that case.

CMakeLists.txt
```diff
@@ -113,6 +117,8 @@ else()
     list(APPEND wxWidgetsComponentsRequired adv)
   endif()
 endif()
+
+message(STATUS "Finding: wxWidgets ver: ${wxWidgetsVersion}")
```
sodevel, line 121: This is superfluous and misleading, `find_package` does already output this, but only if it actually does search the package.

PR #752, opened 2022-08-12
changed build process to use wxWidgets3.2 (on Windows and macOS)

Konrni:
changed README.md and CMakeLists.txt to use wxWidgets3.2 by default.
Added WXFB_WXWIDGETS_ENABLE_31 as fallback option in CMakeLists.txt.

Tested and worked on Windows 10 with MINGW64. 
It looks like ubuntu is still on wxWidgets3.0, so no change there.
Fedora's packages are still for wxWidgets3.1.5. new Flag "WXFB_WXWIDGETS_ENABLE_31" used.
macOS should be fine, since brew's default version is 3.2. [wxWidgets on brew](https://formulae.brew.sh/formula/wxwidgets )

I didn't explicitly test Ubuntu, Fedora or macOS!


sodevel:
I'm sorry, but this PR is mostly superfluous.

The wxWidgets version requested in the CMake files is not required but a lower bound, in the common case you have only one version installed, even with the requested version 3.1.2, it will also use a present 3.2.0 version. The option WXFB_WXWIDGETS_ENABLE_30 was only added to prevent to use a 3.0.x version automatically because these versions are known to cause issues with wxFB. This is not the case for 3.1.x versions, these can be used safely. Also, wxWidgets 3.2 is not a development version, it is a stable version, like 3.0.

Since wxFB does not require wxWidgets 3.2 currently, nothing needs to be changed to build against wxWidgets 3.2. You only need to update your build environment. Instead of just changing the documentation, it is much more interesting to change the GitHub CI Workflows to use wxWidgets 3.2 where available and necessary.

Issue #757, opened 2022-09-04
Codegen crash when a menu is attached to an AUI toolbar tool


davidgiven:
This is with 3.10.1.0, the Debian x64 version.

Adding a menu to a tool in a AUI toolbar causes a codegen crash for C++, Python or PHP. XRC is fine, because it's not doing anything, and Lua is fine, because it doesn't try to do codegen for AUI toolbars.

A minimal example is attached --- just load this and go to the C++ codegen tab.

[minimal.zip](https://github.com/wxFormBuilder/wxFormBuilder/files/9485616/minimal.zip)

davidgiven:
Stacktrace:

```
#0  0x000056217a085a75 in TemplateParser::ParseWxParent() ()
#1  0x000056217a0868fa in TemplateParser::ParseTemplate() ()
#2  0x000056217a088803 in TemplateParser::ParseIfParentTypeEqual() ()
#3  0x000056217a086440 in TemplateParser::ParseMacro() ()
#4  0x000056217a0868fa in TemplateParser::ParseTemplate() ()
#5  0x000056217a086cf2 in TemplateParser::ParseIfEqual() ()
#6  0x000056217a086460 in TemplateParser::ParseMacro() ()
#7  0x000056217a0868fa in TemplateParser::ParseTemplate() ()
#8  0x000056217a0884a2 in TemplateParser::ParseIfNotNull() ()
#9  0x000056217a0868fa in TemplateParser::ParseTemplate() ()
#10 0x000056217a095282 in CppCodeGenerator::GenDefinedEventHandlers(std::shared_ptr<ObjectInfo>, std::shared_ptr<ObjectBase>) ()
#11 0x000056217a09588b in CppCodeGenerator::GetGenEventHandlers(std::shared_ptr<ObjectBase>) ()
#12 0x000056217a0957a1 in CppCodeGenerator::GetGenEventHandlers(std::shared_ptr<ObjectBase>) ()
#13 0x000056217a0957a1 in CppCodeGenerator::GetGenEventHandlers(std::shared_ptr<ObjectBase>) ()
#14 0x000056217a0957a1 in CppCodeGenerator::GetGenEventHandlers(std::shared_ptr<ObjectBase>) ()
#15 0x000056217a0957a1 in CppCodeGenerator::GetGenEventHandlers(std::shared_ptr<ObjectBase>) ()
[...]

PR #758, opened 2022-09-04
Prevent codegen crash when creating drop-down menus.

davidgiven:
This was caused by what looks like bad cut-and-paste in the templates --- the parameter to a #wxparent was removed without removing the #wxparent too. It'd be good if this was more robust.

I've verified that the C++ generated code works, but the PHP and Python was done blindly. At least the codegen doesn't crash any more.

Fixes: #757

### DIFF
--- a/plugins/common/xml/menutoolbar.cppcode
+++ b/plugins/common/xml/menutoolbar.cppcode
@@ -104,7 +104,7 @@ Written by
               #nl wxRect rect = #wxparent $name->GetToolRect( event.GetId() );
               #nl wxPoint pt = #wxparent $name->ClientToScreen( rect.GetBottomLeft() );
               #nl pt = ScreenToClient( pt );
-              #nl #wxparent PopupMenu( $name, pt );
+              #nl #wxparent $name->PopupMenu( $name, pt );
               #nl #wxparent $name->SetToolSticky( event.GetId(), false ); #unindent
             #nl } #unindent
           #nl }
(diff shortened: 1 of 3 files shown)


sodevel (changes requested):
Thanks for finding this, your explanation of the error is not completely correct though.

The `#wxparent` macro resolves a property of the parent wxWidgets object, this requires to specify a property after the macro. And this is wrong in these templates, they don't specify a property but something else, this probably leads to a nullpointer access causing the crash.

Your fix for the C++ side is correct, the fixes for the other languages not, see the followìng review comments. Can you apply these changes and force push this branch? If you do, you will get all the credit when i merge this.

davidgiven:
Reviews addressed --- I think; did you get the Python and PHP recommendations backwards? PTAL.

sodevel:
I am not sure if i understand your question, but no, i did not look up the API for Python or PHP. I looked at the surrounding code and assumed the API is the same like for C++, call the PopupMenu() method of the toolbar with the menu, and this is what my requested changes do. I cannot test the generated code for these languages.

davidgiven:
I mean, your suggestion for the Python file had `@$this->PopupMenu` in it, which is what I'd expect for PHP, and the suggestion for the PHP code had `self.PopupMenu` in it, which is what I'd expect for Python... But I don't work in either of these, so can't test it either!

sodevel:
Oh, my bad, now i see it, i mixed up the boxes during copy-paste. You did the right thing.

sodevel (approved):

davidgiven:
Excellent, thank you!

davidgiven:
I'm actually not at all sure the C++ template is correct. The code fragment in question is emitted into the .h file, not the .cpp file, which seems weird; and also somehow the event handler that handles the menu dropdown is blocking any event handlers connected to the button itself, with the result that in the generated program, clicks on the button don't fire the event (but the dropdown arrow works).

I'll continue to fiddle with it.

PR #781, opened 2023-06-19
Bug Fix: wxSpinCtrlDouble XRC property min and max can't have digits

stirwl:


sodevel:
Thanks for spotting this. These import-export parts always come in pairs, you have fixed the xfb->xrc direction but not the xrc->xfb direction directly below. Can you fix that section as well? Otherwise i will do it myself in the next days.

stirwl:
Thank you, rushed too fast and didn't see it.
Import XRC direction also changed.

PR #788, opened 2023-09-09
Add required Keywords to the desktop file.

melizasw:
Debian requires that desktop files have a "Keywords" entry as described here: https://wiki.gnome.org/Initiatives/GnomeGoals/DesktopFileKeywords

This patch add some keywords that I thought were appropriate. Please modify them as you see fit. Thanks.


sodevel:
I'm not that familiar with the linux desktop environment and simply followed the freedesktop.org documentation where this field is optional. This PR is looking good to me, i'm merging it as is, thanks.

melizasw:
Thanks. I don't know how important it is or why Debian decided to raise a flag when it is missing. Some developers want Lintian as clean as possible before they will sponsor a package so this will help even if it seems a bit pedantic.

Here is the info on the warning that Lintian throws in case anyone is interested: https://lintian.debian.org/tags/desktop-entry-lacks-keywords-entry

PR #789, opened 2023-09-11
fix multiple timers callback bug

cedricboudinet:
Hello
There is an issue when multiple timers are declared in wxFormBuilder. All the timers launches the same callback when using the connect method.
Here is a fix for python and cpp.

When using an event table, one has to create manually the IDs (like for controls) to avoid such an issue.
Regards
Cedric

### DIFF
--- a/plugins/additional/xml/additional.cppcode
+++ b/plugins/additional/xml/additional.cppcode
@@ -378,7 +378,7 @@ Written by
   <templates class="wxTimer">
     <template name="include">@#include &lt;wx/timer.h&gt;</template>
     <template name="declaration">#class @ $name;</template>
-    <template name="construction">$name.SetOwner( #wxparent $name, $id );</template>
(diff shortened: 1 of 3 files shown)


sodevel:
I understand what causes the issue, the current code uses wxID_ANY to bind the handlers, and this matches everything. I see that your PR prevents this by actually querying the resulting wxWindow ID, but i am not sure if this is enough. A few lines above, the SetOwner() call still uses wxID_ANY and according to the documentation, this will be used as ID for the generated events. However, the documentation does not say what does actually happen if you use wxID_ANY for it.

Shouldn't the SetOwner() call use the same ID instead?

cedricboudinet:
I guess you are right. I got a look to the wxTimer source code: calling SetOwner with wxID_ANY will change the timer ID.
I'll update the code.

PR #791, opened 2023-10-13
Fix an issue that generaged XRC has no chance to escape '&' in button label

stirwl:
When set 'My && Button' as a button label to expect the button text be "My & Button", '&' will be converted to '_'.
So that the button label in exported XRC will be 'My __ Button'.

### DIFF
--- a/sdk/plugin_interface/xrcconv.cpp
+++ b/sdk/plugin_interface/xrcconv.cpp
@@ -56,9 +56,6 @@ static wxString StringToXrcText(const wxString& str)
             case '_':
                 result += "__";
                 break;
-            case '&':
-                result += "_";
-                break;
             default:
                 result += ch;
                 break;


sodevel:
Sorry for the late reply, but i have been quite busy with other tasks.

Apparently there is a design flaw in the XRC specification, the escape mechanism is not fully reversible. I have quite a bad feeling about disabling such an essential part and rely on the XML entity processing as last resort to produce valid XML. wxWidgets itself must be affected by this problem, i investigated a bit and found only this pretty old issue https://github.com/wxWidgets/wxWidgets/issues/18033 but i think it applies to your problem. It seems that wxWidgets solves this by conditionally not perfoming some escapes like here https://github.com/wxWidgets/wxWidgets/commit/689704119a3a12110bf8842c76c6298c5e4d9d1d.

I don't know if this is still the correct way to do it and how difficult it would be to implement this for wxFB, but i think we should chose a solution that is compatible with wxWidgets tooling.

PR #792, opened 2023-10-27
adding cpack directives for debian

cedricboudinet:
Hello
I added a piece of code to CMakeLists.txt so that one can easily generate an installer (tested with debian).


sodevel:
The last time i checked (well, i guess 2 years ago) it was quite cumbersome to get a DEB package with all metadata and stuff correctly setup by using CPack, you had to set a lot of variables to pass the required information. So i somehow doubt that setting this single variable is enough to get a proper DEB package.

For that reason i have setup the debian metadata manually, it is available under `data/packaging/linux/debian`. The GitHub CI simply copies this to the root directory and starts the package builder.

cedricboudinet:
A lot of improvements have been made to cpack and some variables are automatically set from cmake setup and some variables are tunable (https://cmake.org/cmake/help/book/mastering-cmake/chapter/Packaging%20With%20CPack.html#cpack-for-debian-packages).
A thing that is a bit tricky is the dependencies which are not detected, and in this case the version number of wx is included in the debian package name.

In fact, I was willing to generate a package for debian bookworm (apt doesn't provices libwx3.0 anymore but switched to 3.2). 
Is it possible to add in the README, the instructions to generate the deb file or to add a 3.10.1 package for ubuntu 23 (which has wx3.2) in the releases section ?

sodevel:
In the current version, the `control` file uses alternations to work with wxWidgets 3.0 and 3.2 giving preference to wxWidgets 3.2. You sould be able to use the included metadata to build for Debian, you might need to modify `changelog` to use Debian code names instead to make SBuild happy.

PR #806, opened 2024-02-10
fix wxFlexGridSizer XRC output of growablerows and growablecols

wsu-cb:
The wxFlexGridSizer growablerows and growablecols
properties are PT_UINTPAIRLIST, not integers

### DIFF
--- a/plugins/layout/layout.cpp
+++ b/plugins/layout/layout.cpp
@@ -320,8 +320,8 @@ public:
         }
         filter.AddProperty(XrcFilter::Type::Integer, "vgap");
         filter.AddProperty(XrcFilter::Type::Integer, "hgap");
-        filter.AddPropertyValue("growablerows", obj->GetPropertyAsString("growablerows"), false);
-        filter.AddPropertyValue("growablecols", obj->GetPropertyAsString("growablecols"), false);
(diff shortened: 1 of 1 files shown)


sodevel:
Thanks for spotting this. I have two remarks about this PR though.

First, there is no reason to store the result of `GetPropertyAsString()` in a local variable, it can just be added directly into the `AddPropertyValue()` call, other code parts do this as well. It would be great if you could change that.

Second, these XRC methods always come in pairs, you fixed the export one, the import one right below is still wrong. It would be nice if you could fix that method as well, otherwise i will do it myself when i have some spare time.

wsu-cb:
Please see if [commit 3d094d](https://github.com/wxFormBuilder/wxFormBuilder/pull/806/commits/3d094d93c00a64c0af34a0055fb7842e2d8a2db2) satisfies your requests.

sodevel:
Thanks for the updates, but now you created an awkward asymmetric solution. On the export path you use `AddPropertyValue()` to prevent the XRC escaping of the value, on the import path you use `AddProperty()` that reverses XRC escaping. That doesn't make sense, you should stick to one approach. These values should contain nothing that gets broken by XRC escaping so i suggest to use the easier `AddProperty()` approach.

PR #808, opened 2024-02-25
Lua: fixed missing parent name for top level box sizer

### LINKED ISSUE #807: Lua code bug when creating a wxBoxSizer
Incorrect Lua code is generated when i add any box sizer onto a new frame.
The problem is these two lines:
```
UI.:SetSizer( UI.gbSizer2 )
UI.:Layout()
```

After i manually add the frame name to UI, it works as intended:
```
UI.MyFrame1:SetSizer( UI.gbSizer2 )
UI.MyFrame1:Layout()
```

Liteb0t:
closes #807


sodevel:
Thanks for this PR, i did some tests on my own and came to the same solution. I just added some comments and the changelog update.

PR #816, opened 2024-05-08
PEP8/pycodestyle pylint fixes

dseichter:
The generated python file has a lot of findings by checking with [pycodestyle](https://pypi.org/project/pycodestyle/) --format=pylint aka [PEP8](https://peps.python.org/pep-0008/).
Eighter you ignore them within the pipeline 

```pycodestyle . --format=pylint --exclude=gui.py```

of the projects or fix it within the generator.

I checked, fixed, added code to fix all the findings within the pythoncode generator, which ends up in no findings anymore, but kept the functionality (tested with my that time converted applications).

Probably (not my decision, only my suggestion) the option to set for Python the setting "indent_with_spaces" to true by default.

### **How I tested:**
I created one project and added ALL components to different panels, frames and dialogs with different layouts.
First try: hundreds of findings

With my PR, 0 (zero) findings left. Now you can remove the ```--exclude=gui.py``` option, running pycodestyle.

My existing application [vatservice-gui](https://github.com/dseichter/vatservice-gui) and the other upcoming applications, I currently migrate from Delphi to Python using wxPython, worked without any further issue.

### CODE REVIEW

**Comment by sodevel** [comment]
Thanks for this PR. First a disclaimer, i know almost zero Python. In general, i don't care much how the generated code looks like, a user should never look at it. Currently, all generators use a similar code style, but this is not a requirement, this happened because new languages got added by making a copy of an existing generator.

In general, i am fine with this PR, however i see two issues. The first one is the removal of newlines. From past experience i know this can be a tricky one, because depending on the made settings, some code fragments get omitted and it might get hard to generate valid code in all different combinations, so sometimes you have to live with excess newlines. Have you somehow verified that all possible paths generate valid code? Especially since i think whitespace is sometimes meaningful in Python.

I did some quick tests and some results did look unintended for me, please see the following code review, also for the second issue.

File: src/codegen/pythoncg.cpp [hunk 1]
```diff
@@ -645,7 +648,7 @@ void PythonCodeGenerator::GenVirtualEventHandlers(const EventVector& events, con
             if (i < (events.size() - 1))
                 m_source->WriteLn();
         }
-        m_source->WriteLn(wxEmptyString);
+        // m_source->WriteLn(wxEmptyString);
```
sodevel, line 651: I think this contributes to this unintended result

```python
    # Virtual event handlers, override them in your derived class

    def onClick(self, event):
        event.Skip()
    # Virtual image path resolution method. Override this in your derived class.
    def wrapme(self, bitmap_path):
        return bitmap_path
```

src/codegen/pythoncg.cpp
```diff
@@ -716,6 +719,10 @@ wxString PythonCodeGenerator::GetCode(PObjectBase obj, wxString name, bool silen
     PythonTemplateParser parser(obj, _template, m_i18n, m_useRelativePath, m_basePath, m_imagePathWrapperFunctionName);
     wxString code = parser.ParseTemplate();
 
+    // add spaces around bitwise or shift operator
+    if (code.Find(wxT(" | ")) == wxNOT_FOUND)
+        code.Replace(wxT("|"), wxT(" | "));
+
```
sodevel, line 725: This is very dangerous, this replacement is done on the fully evaluated code template without knowing and respecting its syntax. This DOES break user code, you can verify this easily by using e.g. a `wxStaticText` with a label `My|Label`, a space padding gets added around the pipe.

sodevel (changes requested):

plugins/additional/xml/additional.pythoncode
```diff
@@ -27,113 +27,113 @@ Python code generation written by
...
-    <template name="evt_connect_OnRichTextStyleSheetChanged">self.$name.Bind( wx.richtext.EVT_RICHTEXT_STYLESHEET_CHANGED, #handler )</template>
-    <template name="evt_connect_OnRichTextStyleSheetReplacing">self.$name.Bind( wx.richtext.EVT_RICHTEXT_STYLESHEET_REPLACING, #handler )</template>
-    <template name="evt_connect_OnRichTextStyleSheetReplaced">self.$name.Bind( wx.richtext.EVT_RICHTEXT_STYLESHEET_REPLACED, #handler )</template>
-    <template name="evt_connect_OnRichTextContentInserted">self.$name.Bind( wx.richtext.EVT_RICHTEXT_CONTENT_INSERTED, #handler )</template>
-    <template name="evt_connect_OnRichTextContentDeleted">$self.$name.Bind( wx.richtext.EVT_RICHTEXT_CONTENT_DELETED, #handler )</template>
+    <template name="evt_connect_OnText">self.$name.Bind(wx.EVT_TEXT, #handler)</template>
+    <template name="evt_connect_OnTextEnter">self.$name.Bind(wx.EVT_TEXT_ENTER, #handler)</template>
+    <template name="evt_connect_OnTextURL">self.$name.Bind(wx.EVT_TEXT_URL, #handler)</template>
+    <template name="evt_connect_OnTextMaxLen">self.$name.Bind(wx.EVT_TEXT_MAXLEN, #handler)</template>ate>
```
sodevel, line 43: There is a syntax error at the end of the line, this was already present before the change, just marking it so it doesn't get lost.

dseichter:
Hello @sodevel,
thank you very much for your time to review my PR. You are deeper in the history of the codegen than myself, so your hints I will respect and rework on my changes.
Additionally to that, I will also extend my test application (with all components on it, with the AUI enabled, to find out more possible issues during the generation of the code.

Some short notes because of the adjustments of the generated code:
I am using since Python 2.7 the PEP8 check (aka pylint) and with version 3.x of Python (2.7 is EOL) there were also a lot of adjustment in how to write code based on the PEP8 style guide. I try to not ignore this style guide, therefore I fight with myself to exclude files 😄 

I need some time, but probably in the next weeks, I have done the changes within your hints and removed the "global" replacements of the | character.

Thanks again a lot,
Daniel

dseichter:
Hello @sodevel,
I would close this PR, because I have a lots of trouble to find the exact locations (C++ is not a language, I am familar with enough). 
What do you think? The workaround is still to ignore the auto generated files in the linters.

sodevel:
Well, i leave this up to you. While i do know C++ i know basically nothing about Python nor do i even have a Python development environment to test generated code. I don't know how difficult it is to fix the output completely but im afraid the newline problem is hard to solve. Since i think excluding generated code from linters and the like is in general a good idea i won't mind if you don't finish this PR.

You don't have to close this PR, you can also leave it lingering here as reference, maybe others have interest to use it as base to fix the generated code, maybe i close it myself during a cleanup process in the future.

PR #853, opened 2024-10-23
Bugfix: some data page widgets' class is exported as unknow in xrc

stirwl:


sodevel:
Thanks for this PR, while it solves the initial issue, it creates new issues as well.

Just adding the minimal conversion code is not sufficient in this case, please see the [XRC File Format](https://docs.wxwidgets.org/latest/overview_xrcformat.html) specification. The `wxDataView*` controls only use the standard window properties (except `wxDataViewTreeCtrl` which also has an `imagelist` property), this means they do not support defined columns. In wxFB you can add columns to `wxDataViewCtrl` and `wxDataViewListCtrl`, now these get exported to XRC as unknown elements, but they must not be exported at all.

Basically the same happens for the `wxPropertyGrid` like elements, but in contrast XRC does support their child elements, just the XRC import / export code is missing or incomplete for these elements. Also these components do support way more attributes than the standard window properties, and although the documentation does not note this explicit, i assume the child elements do not support standard window properties.

The `wxTreeListCtrl` doesn't seem to be supported by XRC, at least it is not mentioned in the documentation.

In the current state this PR is too incomplete to get merged.

stirwl:
Thank you for the information.
Got that treat wxDataViewListCtrl's XRC class as unknow can prevent its child elements being added.

So, XRC specification info should be added to fbp data, or prevent wxFB add child for wxDataViewListCtrl.
Considering wxDataViewListCtrl indeed has child, maybe XRC will include it later.
How about add a field to fbp to indicate wheter an element should have child?

sodevel:
> Got that treat wxDataViewListCtrl's XRC class as unknow can prevent its child elements being added.

This is a side effect of how the XRC conversion works, if the conversion function is not implemented, that `unknown` element is created and the conversion does not step down the chain any more. Since you now implemented the function, the conversion steps down further and hits other elements without a conversion function and hence outputs the `unknown` element for these elements. The issue here his, currently there is not really a solution to output nothing in a conversion function (if you return `nullptr`, the `unknown` element is created, if you return the input element unmodified, it should be invalid to be added to the XML), i need to investigate to find a simple solution that preferrably does not again break the plugin ABI/API.

> So, XRC specification info should be added to fbp data, or prevent wxFB add child for wxDataViewListCtrl. Considering wxDataViewListCtrl indeed has child, maybe XRC will include it later. How about add a field to fbp to indicate wheter an element should have child?

No, the XRC conversion is completely handled by the conversion functions, there is no need to extend the wxFB data model which does already model what children elements can have, here is just a difference between what wxFB supports and what XRC supports.

sodevel:
> The issue here his, currently there is not really a solution to output nothing in a conversion function (if you return `nullptr`, the `unknown` element is created, if you return the input element unmodified, it should be invalid to be added to the XML), i need to investigate to find a simple solution that preferrably does not again break the plugin ABI/API.

I just pushed a commit that solves this problem, returning the input element unmodified (more specifically, keeping its empty name) omits it from the XRC export (or import). This allows to easily drop the columns for the XRC output.

Issue #724, opened 2022-01-25
XRC error: 766: unknown style flag "wxTE_CENTER"


IamSierraCharlie:
Hi,  Using your wxFormBuilder (very helpful app BTW!!) application in Ubuntu Linux 20.04 and have come across this error:
![image](https://user-images.githubusercontent.com/10386637/150950353-0555f255-c10b-43b0-b0c5-29eac553bcce.png)
If I go to the created .xrc file and change the spelling of "CENTER" to "CENTRE", it seems to resolve the issue.

Is this an issue for anyone else or is there something simple that I can adjust to fix this?  I really prefer not to change the xrc file each time I create it.  I guess it depends where live as to how you might spell CENTER/CENTRE.... I hail from Australia so the latter is the way they do it over here :-) 

wxFormBuilder info
version is 3.10.1-39-g978478d5
![image](https://user-images.githubusercontent.com/10386637/150951076-732dc982-c990-4027-a3e6-06be68f39c6f.png)

**wxpython build info:**
wxpython 4.0.4 py37hc99224d_0
-----------------------------
file name   : wxpython-4.0.4-py37hc99224d_0.conda
name        : wxpython
version     : 4.0.4
build string: py37hc99224d_0
build number: 0
channel     : https://repo.anaconda.com/pkgs/main/linux-64

many thanks for your time
Steve

IamSierraCharlie:
I did just find this as a pull request:
https://github.com/wxFormBuilder/wxFormBuilder/pull/449
will being on a different version of wxpython resolve this by chance?

sodevel:
What is actually causing this error message? wxFormbuilder? Your own application? When does it happen?

I am only using C++ and know almost nothing about wxPython but especially nothing about their versioning. These flags got renamed a long time ago in the C++ library and i just renamed them in wxFB too. These changes have happened in some wxWidgets 3.1.x version, i don't know what wxPython version you need to use such a wxWidgets version.

IamSierraCharlie:
I use wxFormBuilder to build my GUI for use in my Python project and the error shows up when I go to run it.
My project uses wxPython 4.1.1.  

![image](https://user-images.githubusercontent.com/10386637/153695217-7cec03ad-c057-4d69-9650-6cfeb0c48453.png)

Its now clear to me that there is a spelling issue.  
[...]

PR #861, opened 2024-11-23
wxTextCtrl has sytle wxTE_CENTRE rather than wxTE_CENTER

stirwl:
#724


stirwl:
I get an error when select some panel in a project. Will check whether it is caused by this commit.
屏幕截图 2024-11-25 094357

stirwl:
That is because a widget has such property:
`<property name="disabled">Load From File; </property>`
It has a bitmap defined before, but removed later. Then "Load From File; " will be left in that property.
Replace `>Load From File; <` to `><` can work.

sodevel:
I checked the wxWidgets sources and discovered that `wxTE_CENTER` does in fact exist, but the XRC handler does only support the `wxTE_CENTRE` variant. Since wxFB uses the `CENTER` variant of other style flags only, i opened a PR https://github.com/wxWidgets/wxWidgets/pull/25118 to add support for the other variant.

Hopefully the wxWidgets folks accept the PR so that this issue will solve itself. Also a note for the future, you cannot simply modify such values that are stored in the wxFB project file and change the project file conversion code for older versions, this will break existing project files. You have to add a new project file version and conversion code for that one.

Issue #869, opened 2025-01-21
Adding  wxTE_MULTILINE flag to wxTextCtrl style triggers an assert failure.


Fat-Zer:
Adding wxTE_MULTILINE flag to wxTextCtrl style triggers an assert failure:

```
ASSERT INFO:
../wxWidgets-3.2.5/src/gtk/textctrl.cpp(910): assert ""IsSingleLine()"" failed in GetEditable(): shouldn't be called for multiline

BACKTRACE:
[1] non-virtual thunk to wxTextCtrl::GetEditable() const
[2] wxTextEntry::SetMaxLength(unsigned long)
[3] TextCtrlComponent::Create(IObject*, wxObject*)
[4] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[5] wxEventHashTable::HandleEvent(wxEvent&, wxEvtHandler*)
[6] wxEvtHandler::TryHereOnly(wxEvent&)
[7] wxEvtHandler::ProcessEventLocally(wxEvent&)
[8] wxEvtHandler::ProcessEvent(wxEvent&)
[9] wxScrollHelperEvtHandler::ProcessEvent(wxEvent&)
[10] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[11] wxEventHashTable::HandleEvent(wxEvent&, wxEvtHandler*)
[12] wxEvtHandler::TryHereOnly(wxEvent&)
[13] wxEvtHandler::ProcessEventLocally(wxEvent&)
[14] wxEvtHandler::ProcessEvent(wxEvent&)
[15] wxWindowBase::TryAfter(wxEvent&)
[16] wxWindowBase::TryAfter(wxEvent&)
[17] wxEvtHandler::SafelyProcessEvent(wxEvent&)
[18] wxPropertyGrid::SendEvent(int, wxPGProperty*, wxVariant*, unsigned int, unsigned int)
[19] wxPropertyGrid::DoPropertyChanged(wxPGProperty*, unsigned int)
[20] wxPropertyGrid::HandleCustomEditorEvent(wxEvent&)
[21] wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&)
[22] wxEventHashTable::HandleEvent(wxEvent&, wxEvtHandler*)
[...]

PR #870, opened 2025-01-22
Fix assertion failure on multiline wxTextCtrl

Fat-Zer:
When attempting to use a wxTextCtrl with wxTE_MULTILINE, it generates an assertion failure. As wxWidgets documentation says[1], in wxGTK SetMaxLength() may only be used with single line text controls.

[1]: https://docs.wxwidgets.org/3.2/classwx_text_entry.html#a5b9dea0d1adeb9cc14309600de6aff50

Closes: https://github.com/wxFormBuilder/wxFormBuilder/issues/869

### DIFF
--- a/plugins/common/common.cpp
+++ b/plugins/common/common.cpp
@@ -552,7 +552,11 @@ public:
           obj->GetPropertyAsSize(_("size")),
           obj->GetPropertyAsInteger(_("style")) | obj->GetPropertyAsInteger(_("window_style")));
 
-        if (!obj->IsPropertyNull(_("maxlength"))) {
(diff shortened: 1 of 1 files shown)


Fat-Zer:
PS: I'm not sure, if there is any need for changes in code generators.

Fat-Zer:
> The wxWidgets documentation says this limitation does also apply to wxQt, but i haven't found a define for that platform in the documentation

I suppose it's `__WXQT__`, isn't it? But AFAIR the Qt backend is semi-finished and it would require much more patches than just this one to make it work.

sodevel:
Thats what i thought too, but i didn't find it in the documentation and didn't bother to check the sources.

PR #888, opened 2025-07-14
Fix startup on wayland

### LINKED ISSUE #854: gtk_widget_set_size_request: assertion 'width >= -1' failed
OS: Arch
wxFormBuilder Version: 4.2.1-0-g80c4cb6
wxWidgets-gtk3 Version: 3.2.6-1

When I start wxformbuilder it hangs forcing you to need to kill the process `kill $pid`. I thought I'd try from the terminal and I received this error printed twice.

Error:
`(wxformbuilder:776325): Gtk-CRITICAL **: 17:51:39.959: gtk_widget_set_size_request: assertion 'width >= -1' failed`

I did find a couple of similar yet different issues referencing `gtk_widget_set_size_request: assertion 'width >= -1`:

1. https://github.com/wxFormBuilder/wxFormBuilder/issues/215
2. https://github.com/wxFormBuilder/wxFormBuilder/issues/237
3. https://github.com/wxWidgets/wxWidgets/issues?q=is%3Aissue+gtk_widget_set_size_request%3A+assertion+%27width+%3E%3D+-1+is%3Aclosed

I found the ones on wxWidgets and included them in case they help.

If I can provide more or better information let me know. Thanks!

schorsch1976:
This let me start wxformbuilder on wayland.

Should close #854

### DIFF
--- a/src/rad/appdata.cpp
+++ b/src/rad/appdata.cpp
@@ -2682,7 +2682,7 @@ void ApplicationData::NotifyProjectSaved()
 void ApplicationData::NotifyProjectRefresh()
 {
     wxFBEvent event(wxEVT_FB_PROJECT_REFRESH);
-    NotifyEvent(event);
+    NotifyEvent(event,true);
 }
 
 


sodevel:
This is a very dangerous change, this event is emitted and used in multiple locations and surrounding code might require this event already beeing processed to be able to continue. Doing some quick testing i already found one case where this change causes an error, moving a toolbar from a frame into a sizer and the reverse.

Since this change fixes the startup for you, i wonder why this change has this effect. I checked the wxFB startup sequence, apparently inside our `OnRun()` an empty project gets created which emits this event, but in this method no event loop should be running, this only gets started by the base implementation that gets executed later. Maybe this is an issue on Wayland and the cause for the crash.

The whole code of the startup sequence is pretty old and quite weird, it also contains a `wxYield()` which should do nothing there. I think a more safe approach is to rework that part so that nothing that emits events gets executed before the event loop is running.

This will take some time and unfortunately i cannot test if this fixes the issues on wayland, maybe for a quick test it is enough to wrap the `AppData()->NewProject();` call in a `CallAfter()` call to check if this fixes the crash on startup under wayland.

schorsch1976:
As i commented on the issue, it appears to me, the startup fails to process an event. The trick with stace slows the startup and the signals from stace can interfer with syscalls. (EAGAIN). As an example see man 2 read.

I got several wx apps running on windows and Linux, none of my apps or other apps I tried from the debian repo shows that startup behavior as wxFB. I think it's related just to wxFB.

Please let me know if I could support that issue :)

https://www.man7.org/linux/man-pages/man2/read.2.html

